PR #20, opened 2025-10-13
Add docs, fix `dictGet()`, add other `uniq` aggs

theory:
Add `doc/clickhouse_fdw.md` and start documenting the extension, starting with `clickhouse_raw_query` and the custom functions and aggregates. Also more thoroughly list the past and present maintainers and copyright holders.

Rename the `clickhouse_pushdown` C function to `clickhouse_func_push_fail` to better describe what it does. Rename its SQL version to `ch_push_agg_text` and add `ch_push_func_text`. Use this latter function in `dictGet()` to more simply raise an error. Also, uncomment the `dictGet()` test to show the function now works.


src/custom_types.c
```diff
@@ -273,10 +273,18 @@ CustomObjectDef *chfdw_check_for_custom_function(Oid funcid)
 					strcpy(entry->custom_name, "argMax");
 				else if (strcmp(proname, "argmin") == 0)
 					strcpy(entry->custom_name, "argMin");
-				else if (strcmp(proname, "dictget") == 0)
-					strcpy(entry->custom_name, "dictGet");
 				else if (strcmp(proname, "uniqexact") == 0)
 					strcpy(entry->custom_name, "uniqExact");
+				else if (strcmp(proname, "uniqcombined") == 0)
```
serprex, line 278: at what point should this use some kind of data structure for lookup?

serprex (approved):

src/custom_types.c
theory: Yeah, I was wondering that myself, especially when I had a bunch of other functions there for a minute. Then I realized we should probably focus on mapping Postgres features to ClickHouse, rather than the other way around, so happy to defer it for now.

PR #26, opened 2025-10-31
Implement BeginInsert/InsertData/EndInsert flow

theory:
Replace the old (and commented-out) code with the new `BeginInsert` / `InsertData` / `EndInsert` pattern in `clickhouse-cpp`. This greatly simplifies the code, as the most of the necessary logic for creating the insert Block object and tracking the state of the insert is now handled by the Client object.

Remove the old logic for unwrapping `LowCardinality` columns and fix `column_append` to properly append to those columns. This preserves traffic over the wire, I'm told. Thanks to @slabko for working out how to make that work again.

Restore the logic for inserting arrays by creating a column of the proper type, then use `column_append()` to add the items to it before appending it to the `ColumnArray`. Thanks to @slabko again for the incantation to create a column of the proper type for array items.

Copy the test from the old FDW, but comment out the bit working with arrays, as it has not yet been updated.

### DIFF
--- a/src/binary.cpp
+++ b/src/binary.cpp
@@ -288,98 +288,69 @@ static Oid get_corr_postgres_type(const TypeRef & type)
 void ch_binary_insert_state_free(void * c)
 {
 	auto * state = (ch_binary_insert_state *)c;
-	if (state->columns)
+	if (state->insert_block)
 	{
-		/* try to send empty block that sets proper ClickHouse state */
-		if (!state->success)
+		/* Finish the insert to set the proper ClickHouse state */
+		Client * client = (Client *)state->conn->client;
+		try
 		{
-			try
-			{
-				Client * client = (Client *)state->conn->client;
-				client->Insert(state->table_name, Block());
-			}
-			catch (const std::exception & e)
-			{
-				// just ignore, next query will fail
-				elog(NOTICE, "pg_clickhouse: could not send empty packet");
-			}
+			client->EndInsert();
 		}
-
-		delete (std::vector<clickhouse::ColumnRef> *)state->columns;
+		catch (const std::exception & e)
+		{
+			// just ignore, next query will fail
+			elog(NOTICE, "pg_clickhouse: could not finish INSERT: - %s", e.what());
+		}
+		delete (Block *)state->insert_block;
 	}
 }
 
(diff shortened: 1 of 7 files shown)


serprex (approved):

src/binary.cpp
```diff
@@ -285,98 +285,60 @@ static Oid get_corr_postgres_type(const TypeRef & type)
...
+		// The Client retursn a `unique_ptr`. Grab the internal pointer to
+		// take ownership of it and manage its memory ourselves.
+		block = client->BeginInsert(std::string(query) + " VALUES").release();
+	}
+	catch (const std::exception & e)
+	{
+		elog(ERROR, "clickhouse_fdw: could prepare insert - %s", e.what());
+	}
+	state->len = block->GetColumnCount();
+	if (state->len == 0) return;
```
serprex: leaks `block`

theory: I _think_ this is fixed now.

theory: Fixed in [b2df9d3](https://github.com/ClickHouse/clickhouse_fdw/compare/de366b115478a02b4e97a5a5afe7df9a829a49b3..b2df9d3ec2298c2c97f193967fd76d9aaccc20a4), thanks.

src/binary.cpp
```diff
@@ -567,25 +531,17 @@ static void column_append(clickhouse::ColumnRef col, Datum val, Oid valtype, boo
...
-						+ col->Type()->GetName()
-					);
-					// auto arrcol = col->As<ColumnArray>();
-
-					// arrcol->OffsetsIncrease(arr->len);
-					// for (size_t i = 0; i < arr->len; i++)
-					// 	column_append(
-					// 		arrcol->Nested(), arr->datums[i], arr->item_type, arr->nulls[i]);
+					auto arrcol = col->AsStrict<ColumnArray>();
+					auto items = arrcol->NewColumn();
```
slabko:
```suggestion
					    auto items = Ch::CreateColumnByType(arrcol->GetType().As<Ch::ArrayType>()->GetItemType()->GetName());
```

theory: Thank you! Fixed in a7b4c7c11aa907db6ea4b2292c5802f651438d68.

PR #27, opened 2025-10-31
Set default host name in `ch_http_connection()`

### LINKED ISSUE #18: clickhouse_raw_query() Panics when not host name
```
david=# select clickhouse_raw_query('SELECT version()', '');
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
The connection to the server was lost. Attempting reset: Failed.
```

Works fine:

```
david=# select clickhouse_raw_query('SELECT version()', 'host=localhost');
 clickhouse_raw_query 
----------------------
 25.9.2.1            +
```

Probably need to set a default host name in `clickhouse_raw_query`.

theory:
Otherwise, calls to `clickhouse_raw_query()` with a connection string that omits the host name causes an error. Using "localhost" as the default to match the default specified in the `CREATE FUNCTION` default. Add an empty connection string argument to a call to `clickhouse_raw_query()` in `tests/sql/http.sql` to test it.

Resolves #18.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -69,6 +69,8 @@ ch_http_connection_t *ch_http_connection(char *host, int port, char *username, c
 	if (!conn->curl)
 		goto cleanup;
 
+	if (!host || !*host) host = "localhost";
+
 	if (!port)
 		port = ch_is_cloud_host(host) ? CLICKHOUSE_TLS_PORT : CLICKHOUSE_PORT;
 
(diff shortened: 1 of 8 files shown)


serprex (approved):
do we also want to fallback to localhost when `!*host`?

PR #33, opened 2025-11-07
Add quantile(), params() for parametric aggregates

theory:
ClickHouse provides a number of "parametric aggregates" which take one or more parameters in a first set of parentheses and then normal arguments in a second. An example is `quantile`, which can be called like so:

    SELECT quantile(0.25)(val) FROM t;

To mimic this feature, implement a `params` data type and function that takes a list of values of any type. Detect when it's the first argument to any aggregates and pass it as ClickHouse parameters before any other arguments. The equivalent of the above query in PostgreSQL is:

    SELECT quantile(params(0.25), val) FROM t;

The pattern is general and can be used for any aggregate functions, but used here only for the new `quantile` and `quantileExact` aggregate functions. Others can be added in the future; all they have to do is use the `clickhouse_func_push_fail` function for their `SFUNC`, with one variant that takes `params` as the first argument and the remaining arguments specified as usual (likely as `"any"` to allow data type enforcement to be pushed down to ClickHouse).

Done by using the new `ch_get_params_function() function to detect `params()` as the first argument to an aggregate function in `deparseAggref()` and emitting it before appending the arguments opening parenthesis.

Document the new functions and explain how to use them.

While at it, simplify the handling of `F_REGEXP_LIKE_TEXT_TEXT` in Postgres 14 and earlier by simply setting it on those versions, rather than using `#define` to omit any code that uses it. This aligns with how the type mapper handles other such data type constants.

Also:

*   Fix an `Assert` for a variadic array to apply to the array, rather than the parent node. *   Also, remove the unused `func_arg` field from the `deparse_expr_cxt` struct and add a comment explaining the `array_as_tuple` field.

### DIFF
--- a/src/deparse.c
+++ b/src/deparse.c
@@ -81,10 +81,9 @@ typedef struct deparse_expr_cxt
 	StringInfo	buf;			/* output buffer to append to */
 	List	  **params_list;	/* exprs that will become remote Params */
 	CustomObjectDef	*func;		/* custom function deparse */
-	void		*func_arg;		/* custom function context args */
 	CHFdwRelationInfo *fpinfo;	/* fdw relation info */
 	bool		interval_op;
-	bool		array_as_tuple;
+	bool		array_as_tuple; /* determines array output format */
 } deparse_expr_cxt;
 
 #define REL_ALIAS_PREFIX	"r"
@@ -3297,6 +3296,7 @@ deparseAggref(Aggref *node, deparse_expr_cxt *context)
 	uint8	brcount = 1;
 	bool	use_variadic;
 	bool 	omit_star;
+	int 	first_arg = 0;
 
 	/* Only basic, non-split aggregation accepted. */
 	Assert(node->aggsplit == AGGSPLIT_SIMPLE);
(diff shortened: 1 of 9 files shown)


src/deparse.c
```diff
@@ -2023,16 +2038,23 @@ deparseArray(Datum arr, deparse_expr_cxt *context)
...
-		appendStringInfoChar(buf, '[');
+	switch (context->array_context) {
+		case FdwArrayContextArray:
+			appendStringInfoChar(buf, '[');
+			break;
+		case FdwArrayContextTuple:
+			appendStringInfoChar(buf, '(');
+			break;
+		case FdwArrayContextList:
+			;
```
serprex:
```suggestion
			break;
```
& later on

serprex (approved):

src/deparse.c
theory: Reverted this bit.

serprex (approved):
it's hard since it's always going to be awkward how to map `(args1)(args2)` to postgres. this way at least makes it clearer there's something magical happening

PR #45, opened 2025-11-13
Propagate aggregate detection to outer node

theory:
The `foreign_expr_walker` function recursively walks the nodes in an SQL query. To detect whether a table column was configured as an `AggregateFunction`, the `T_Aggref` node checks to see if the next node sets `found_AggregateFunction`. For most aggregate functions, the `T_Var` node is next, so `T_Aggref` sees it and sets it properly.

However, sometimes the `T_Var` node is *not* the next node after a `T_Aggref` node. For ordered set aggregates such as `percentile_cont`, there can be a `T_FuncExpr` node, a `T_List`, and *then* a `T_Var` node. But the setting of `found_AggregateFunction` was not propagated from the recursive call, so it would be lost. This resulted in `percentile_cont`, which maps to ClickHouse's `quantile` function, not getting the `Merge` suffix for a `AggregateFunction(quantile)` column.

So set the outer (previous) context's `found_AggregateFunction` to true if the current inner context has set it to true. This allows its value to propagate back to `T_Aggref` and be detected (and then unset). As a result, it properly appends `Merge` to `quantile`, thus properly executing against an `AggregateFunction(quantile)` column.

Add a test case to `test/sql/engines.sql` demonstrating this functionality, including the remote SQL converting `percentile_cont` to `quantileMerge` to query a `AggregateFunction(quantile)` column. Add another test that includes following nodes, including another aggregate function, to ensure that a true `found_AggregateFunction` value only propagates to the previous `T_Aggref`, not any subsequent `T_Aggref` nodes.

While at it, tell the ClickHouse and Postgres workflows not to fail all jobs in a matrix if one fails. This will make it easier to identify which versions fail and which do not.

### DIFF
--- a/.github/workflows/clickhouse.yml
+++ b/.github/workflows/clickhouse.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   build:
     strategy:
+      fail-fast: false
       matrix:
         # Test latest version and all LTS releases.
         # .github/ubuntu/ch-versions --lts --latest-stable --earliest 22  --prefix '          - ' | pbcopy
--- a/.github/workflows/postgres.yml
+++ b/.github/workflows/postgres.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   build:
     strategy:
+      fail-fast: false
       matrix:
         pg: [18, 17, 16, 15, 14, 13]
         os:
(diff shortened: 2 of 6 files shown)


serprex (approved):
that was a smaller change than expected

PR #52, opened 2025-11-19
Add support for the `Decimal` data type

theory:
The mappings already work for the HTTP engine, although its inserts were limited to `float8`, now fixed to use proper numeric formatting.

Add the mappings for the binary engine. Use strings to convert between the Postgres and ClickHouse formats to as to avoid differences in their implementations: ClickHouse stores data as `Int32`, `Int64`, or `Int128`, while Postgres...well, I have no idea what the internals are, but using the `numeric_in` and `numeric_out` functions avoid all issues.

Conversion from `Decimal` values returned from ClickHouse requires converting its internal `Int128` into a string. The clickhouse-cpp library doesn't provide this functionality, [yet], so implement it here.

Unfortunately, this approach doesn't work with clickhouse-cpp compiled into a dynamic library, presumably because it doesn't also compile and install the `absl` dynamic library. So make the build static by default on all platforms for now.

Add a test that covers both the HTTP and binary engines.

  [yet]: https://github.com/ClickHouse/clickhouse-cpp/pull/451

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,9 +32,9 @@ CH_CPP_FLAGS = -D CMAKE_BUILD_TYPE=Release -D WITH_OPENSSL=ON
 
 # Build static on Darwin by default.
 ifndef ($(CH_BUILD))
-ifeq ($(OS),Darwin)
+# ifeq ($(OS),Darwin)
 	CH_BUILD = static
-endif
+# endif
 endif
 
 # Are we statically compiling clickhouse-cpp into the extension or no?
--- a/src/binary.cpp
+++ b/src/binary.cpp
@@ -1,3 +1,4 @@
+#include <sstream>
 #include <iostream>
 #include <cassert>
 #include <stdexcept>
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -32,9 +32,9 @@ CH_CPP_FLAGS = -D CMAKE_BUILD_TYPE=Release -D WITH_OPENSSL=ON
 
 # Build static on Darwin by default.
 ifndef ($(CH_BUILD))
-ifeq ($(OS),Darwin)
+# ifeq ($(OS),Darwin)
```
serprex, line 35: what's this?

theory: I'm forcing it to make a static build by default. It was defaulting to a dynamic build on Linux, but it started failing with this change, presumably because building clickhouse-cpp with `-D BUILD_SHARED_LIBS` needs to also build and install the absl library.

theory: I opted not to change it to

```
CH_BUILD ?= static
```

Because I hope to restore the dynamic build before long.

serprex (approved):

PR #63, opened 2025-12-03
Add session_settings GUC and chfdw_parse_options

theory:
Port libpq's `conninfo_parse()` to `chfdw_parse_options()` with an argument to optionally require comma-delimited rather than space-delimited key/value pairs. This allows it to be used for connection string parsing and for parsing options akin to foreign server, foreign table, and user mapping comma-delimited options.

Refactor `connstring_parse()` to use `chfdw_parse_options()`.

Register a global configuration setting, not yet used, named `pg_clickhouse.session_settings`. Make its default value `join_use_nulls=1` and use `chfdw_parse_options()` to validate it.

Add a tests named `gucs` and use it for now to test the success and failure of option parsing. It contains foreign objects for future commits to use to test propagation of `pg_clickhouse.session_settings` to ClickHouse.

Other notes:

*   Move `_PG_init` to `option.c` to match the organization used by postgres_fdw
*   Only use `PGDLLEXPORT` on Postgres 15 and earlier, as it's no longer needed for later versions

### DIFF
--- a/src/connection.c
+++ b/src/connection.c
@@ -204,115 +204,22 @@ chfdw_inval_callback(Datum arg, int cacheid, uint32 hashvalue)
 ch_connection_details *
 connstring_parse(const char *connstring)
 {
-	char	   *pname;
-	char	   *pval;
-	char	   *buf;
-	char	   *cp;
-	char	   *cp2;
+	ListCell   *lc;
+	List	   *options = chfdw_parse_options(connstring, false);
 	ch_connection_details *details = palloc0(sizeof(ch_connection_details));
 
-	/* Need a modifiable copy of the input string */
-	if ((buf = pstrdup(connstring)) == NULL)
-		return NULL;
+	if (options == NIL)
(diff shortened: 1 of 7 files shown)


src/connection.c
```diff
@@ -204,135 +204,47 @@ chfdw_inval_callback(Datum arg, int cacheid, uint32 hashvalue)
...
-				{
-					*cp2 = '\0';
-					cp++;
-					break;
-				}
-				*cp2++ = *cp++;
-			}
-		}
+		DefElem    *elem = (DefElem *) lfirst(lc);
+		char	   *pname = pstrdup(elem->defname);
```
serprex: why dup defname? `pname` doesn't appear to escape

theory: I guess it doesn't have to be duped because `List` is allocated in the Postgres memory context? Previously this function required dupes, but I just removed these `pstrdup`s and I get no crash. So, good?

PR #66, opened 2025-12-04
Improve unsigned integer support

theory:
Add a regression test that fetches the min and max integer and float values from ClickHouse. Fix a few issues they showed:

*   Teach the binary engine to use an `int32` instead of a `int16` when converting from UInt16
*   Improve the error message raised by the binary engine for `UInt64` values greater than the `BIGINT` max to align with how Postgres itself displays the error

The handling of Float32 and Float64 values changed for the better in ClickHouse 23, so add `test/expected/import_schema_2.out` to cover ClickHouse 22.

While at it, remove `strtoint32`, originally copied from the PostgreSQL source, and use the `pg_strtoint32` function directly.

### DIFF
--- a/src/binary.cpp
+++ b/src/binary.cpp
@@ -697,8 +697,11 @@ nested_col:
 		break;
 		case Type::Code::UInt64: {
 			uint64 val = col->As<ColumnUInt64>()->At(row);
+			/* XXX Consider using, e.g., https://pgxn.org/dist/uint128. */
 			if (val > LONG_MAX)
-				throw std::overflow_error("pg_clickhouse: int64 overflow");
+				throw std::overflow_error(
+					"value " + std::to_string(val) + " is out of range of bigint"
+				);
 
 			ret = Int64GetDatum((int64)val);
 			*valtype = INT8OID;
--- a/test/expected/import_schema.out
+++ b/test/expected/import_schema.out
@@ -656,6 +656,73 @@ EXPLAIN VERBOSE SELECT * FROM clickhouse.custom_option;
    Remote SQL: SELECT a FROM import_test.custom_option
 (3 rows)
 
+-- check overflows.
+SELECT clickhouse_raw_query($$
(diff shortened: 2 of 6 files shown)


test/expected/import_schema.out
```diff
@@ -656,6 +656,66 @@ EXPLAIN VERBOSE SELECT * FROM clickhouse.custom_option;
...
+ 
+(1 row)
+
+SELECT * FROM clickhouse_bin.ints
+WHERE c1 IN (127, -128)
+ORDER BY c1;
+  c1  |   c2   |     c3      |          c4          | c5  |  c6   |     c7     | c8 |      c9       |           c10           
+------+--------+-------------+----------------------+-----+-------+------------+----+---------------+-------------------------
+ -128 | -32768 | -2147483648 | -9223372036854775808 |   0 |     0 |          0 |  0 | 1.1754942e-38 | 2.2250738585072014e-308
+  127 |  32767 |  2147483647 |  9223372036854775807 | 255 | 65535 | 4294967295 | -1 | 3.4028233e+38 | 1.7976931348623157e+308
```
theory: That `-1` for column `c8` is the overflow value for `18446744073709551615` displayed as a `BIGINT`.

src/binary.cpp
```diff
@@ -696,11 +697,9 @@ static Datum make_datum(clickhouse::ColumnRef col, size_t row, Oid * valtype, bo
 		}
 		break;
 		case Type::Code::UInt64: {
+			/* DatumUint64 ok, but INT8OID overflows when val > LONG_MAX. */
 			uint64 val = col->As<ColumnUInt64>()->At(row);
-			if (val > LONG_MAX)
-				throw std::overflow_error("pg_clickhouse: int64 overflow");
-
-			ret = Int64GetDatum((int64)val);
+			ret = DatumGetUInt64(val);
```
theory: I think I'm changing my mind about this; it should not return a negative value when it's greater than `LONG_MAX`, it should raise an error. Maybe if we find we really need to support UInt64 (not to mention 128), we should depend on [pg-uint128](https://pgxn.org/dist/uint128/) or a similar project.

PR #68, opened 2025-12-06
Set PG_MODULE_MAGIC_EXT version to full `x.y.z`

theory:
The plan is to make `x.y.0` releases that increment `.y` only when there are SQL changes, and to release `x.y.1+` for library-only changes. This will allow everyone to get all the library-only bug fixes without having to run `ALTER EXTENSION UPDATE`.

But while the version listed in the catalog will be `x.y`, it makes sense to properly version the library with `x.y.z`. So change the value assigned to `PG_MODULE_MAGIC_EXT.version` to the full release `x.y.z`, so that users can find the full version, at least on Postgres 18 and higher.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -114,7 +114,7 @@ sql/$(EXTENSION)--$(EXTVERSION).sql: sql/$(EXTENSION).sql
 
 # Versioned source file.
 src/fdw.c: src/fdw.c.in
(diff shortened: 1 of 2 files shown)


serprex (approved):
long term it'd be nice to get rid of `.c.in` & instead have a rule that generates a header that defines __VERSION__ as macro

PR #69, opened 2025-12-08
Add `toUInt*()` pushdown cast functions

theory:
One can use `CAST(x AS integer)` in PostgreSQL queries to push down casts to signed integers, but not unsigned integers. Same goes for `smallint` and `bigint`. But one cannot push down a cast to a signed integer, because PostgreSQL has no signed integer types.

So add custom `toUInt*()` functions to pg_clickhouse that push down  to their ClickHouse equivalents. As with other pushdown-only functions, they raise an error when they're not pushed down. Add a new test for these functions, as well as test cases to ensure errors on pushdown failure.

While at it, refactor the processing of custom function names to use a lookup table and a function to perform the lookup. Currently it simply performs a string comparison on each item in the table, but as the table grows we can refactor it to optimize lookups by, say, switching on the first character of the string.

Also add the macros `STR_EQUAL` and `STR_STARTS_WITH` to `custom_types.c` to add a bit more semantic meaning to their uses. We'll likely want to use them elsewhere (in fact `STR_STARTS_WITH` is not yet used), but this is a start.

Use `STR_EQUAL` throughout the file, in particular replacing redundant calls to `NameStr(procform->proname)` with just `proname`, which is already extracted.

Also: `make indent` reformats `PG_MODULE_MAGIC_EXT`.


src/custom_types.c
```diff
@@ -51,6 +51,8 @@
 #define F_REGEXP_LIKE_TEXT_TEXT 6263
 #endif
 
+#define STR_STARTS_WITH(str, sub) strncmp(str, sub, strlen(sub)) == 0
+#define STR_EQUAL(a, b) strcmp(a, b) == 0
```
serprex, line 55: I'd prefer these to be `static inline` functions, but that might just be my preprocessor aversion

serprex (approved):

PR #94, opened 2025-12-13
Add kv_list and refactor session_settings with it

theory:
Compare with #95.

A new structure, kv_list, represents a list of key/value pairs. It does not use an array of structs for each pair, because a custom GUC value must be allocated as a single block,

The `kv_list` constructor takes a Postgres `List` argument, iterates over the keys and values, and then allocates a block of memory for them (with options to use `palloc`, `malloc`, or `guc_malloc` (which falls back on `malloc` prior to Postgrs 16)). It then iterates over them again to pack them into the memory space beyond the `kv_list` struct itself. The design follows the precedent of the `ConvertTimeZoneAbbrevs()` function in Postgres's `datetime.c`.

Use it in `chfdw_check_settings_guc()` to create a `kv_list` from the `List` returned by `chfdw_parse_options()` and assign it to `**extra`. Postgres hands this pointer to `chfdw_settings_assign_hook()`, where we assign it to the `ch_session_settings_list` global, which can then be fetched via `chfdw_get_session_settings()`.

Of course this would be a tricky interface to iterate manually, so also add another struct, `kv_iter`, that handles the logic of iterating through the keys and values in a `kv_list`. Since it sets up the values for each iteration of the loop, the logic using it is simpler than the previous use of `List`, as seen in the http and binary engines.

While at it, cChange `chfdw_parse_options()` to use `pstrdup()` instead of `strdup`, since its memory is handled by a Postgres memory context (and surely was leaking before).

### DIFF
--- a/src/include/kv_list.h
+++ b/src/include/kv_list.h
@@ -1,0 +1,60 @@
+#ifndef PG_CLICKHOUSE_KV_LIST_H
+#define PG_CLICKHOUSE_KV_LIST_H
+
+#include <stdbool.h>
+#include "postgres.h"
+#include "nodes/pathnodes.h"
+
+/*
+ * A simple data structure with a list of key/value string pairs. Use
+ * new_kv_list_from_list() to create.
+ */
+typedef struct kv_list
+{
+	int			length;
+	/* key/value char * pairs follow the length field. */
+	char data[];
+}			kv_list;
+
+/*
+ * Iterator for a kv_list. Use new_kv_iter() to create.
+ *
+ *     for (kv_iter iter = new_kv_iter(kv); !kv_iter_done(&iter); kv_iter_next(&iter))
+ *     {
+ *         printf("%i, %s => %s\n", iter.num, iter.name, iter.value);
+ *     }
+ */
+typedef struct kv_iter {
+    int togo;
+    char * name;
+    char * value;
+} kv_iter;
+
+/*
+ * Defines the allocator to use when creating a new kv_list.
+ * kv_pair_guc_malloc is the same as kv_pair_malloc on Postgres 15 and
+ * earlier, so be sure to free() the memory on those versions.
+ */
+enum kv_pair_alloc
+{
+	kv_pair_guc_malloc,
+	kv_pair_malloc,
+	kv_pair_palloc,
+};
+
+/*
+ * Create a new kv_list from a PostgreSQL List of DefElem. Allocate the memory
+ * using the specified allocator.
+*/
+kv_list * new_kv_list_from_pg_list(List * list, int allocate);
+
+/* Create a new kv_iter for a key_pairs. */
+kv_iter new_kv_iter(const kv_list * ns);
+
+/* Iterate to the next item. Returns false if there are no items. */
+bool kv_iter_next(kv_iter * state);
+
+/* Retruns true if iteration by kv_iter is complete. */
+bool kv_iter_done(kv_iter * state);
+
+#endif							/* PG_CLICKHOUSE_KV_LIST_H */
--- a/src/kv_list.c
+++ b/src/kv_list.c
@@ -1,0 +1,117 @@
+#include <stddef.h>
+#include "kv_list.h"
+#include "utils/elog.h"
+#include "nodes/pathnodes.h"
+#include "utils/guc.h"
+
+static kv_list * allocate_list(size_t size, int allocate)
+{
+	kv_list    *pairs;
+
+	switch (allocate)
+	{
+		case kv_pair_guc_malloc:
+			/* Fall through to malloc prior to Postgres 16. */
+#if PG_VERSION_NUM >= 160000
+			pairs = guc_malloc(ERROR, size);
+			break;
+#endif
+		case kv_pair_malloc:
+			pairs = malloc(size);
+			break;
+		case kv_pair_palloc:
+			pairs = palloc(size);
+			break;
+		default:
+			ereport(ERROR,
+					(errcode(ERRCODE_FDW_ERROR),
+					 errmsg("unknown kv_pair_alloc %i", allocate)));
+	}
+
+	if (!pairs)
+		ereport(ERROR,
+				(errcode(ERRCODE_FDW_OUT_OF_MEMORY),
+				 errmsg("out of memory")));
+
+	return pairs;
+}
+
+extern kv_list * new_kv_list_from_pg_list(List * list, int allocate)
+{
+	ListCell   *lc;
+	DefElem    *elem;
+	kv_list    *pairs;
+	size_t		kv_size = 0;
+	size_t		ck_size = 0;
+
+	/* Count up space for dynamic key/value pairs. */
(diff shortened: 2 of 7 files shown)


src/include/kv_list.h
```diff
@@ -0,0 +1,59 @@
...
+#include "nodes/pathnodes.h"
+
+/*
+ * A simple data structure with a list of key/value string pairs. Use
+ * new_kv_list_from_list() to create.
+ */
+typedef struct kv_list
+{
+	int			length;
+	/* key/value char * pairs follow the length field. */
```
serprex, line 15: if we embrace C99 in extension, can use [flexible array](https://en.wikipedia.org/wiki/Flexible_array_member)

src/kv_list.c
```diff
@@ -0,0 +1,118 @@
...
+{
+	const		size_t kv_size = kv_list_struct_size;
+	char	   *name;
+
+	/* The list may be NULL or empty. */
+	if (!ns || ns->length < 1)
+		return (kv_iter)
+	{
+		0, "", ""
+	};
```
serprex, line 90:
```suggestion
		return kv_iter{};
```

src/include/kv_list.h
theory: Added `char data[];` in 5825410, which is nice for documentation but doesn't actually do anything, since each key and value is variable in size, so it wouldn't be possible to have something like `kv_pair pairs[]`, because then much memory does each one need?

src/kv_list.c
theory: I don't think that will work:

```
type name is not allowed C/C++(254)
```

But even if it did, it would be a problem because `length` **must** be set to zero. And without assigning it, what is the value?

This seems to work, though:

```c
return (kv_iter) {0};
```

theory: Of course then `make indent` reformats it as

```c
		return (kv_iter)
	{
		0
	};
```

😭

src/kv_list.c
```diff
@@ -0,0 +1,118 @@
+#include <stddef.h>
+#include "kv_list.h"
+#include "utils/elog.h"
+#include "nodes/pathnodes.h"
+#include "utils/guc.h"
+
+static const size_t kv_list_struct_size = sizeof(kv_list);
+
```
serprex:
```suggestion
```
I'd prefer using `sizeof(kv_list)` throughout

src/kv_list.c
```diff
@@ -0,0 +1,117 @@
...
+	};
+}
+
+extern bool
+kv_iter_next(kv_iter * iter)
+{
+	if (iter->togo == 0)
+		return false;
+	iter->togo--;
+	iter->name += 2 + strlen(iter->name) + strlen(iter->value);
```
serprex: can't we optimize out the strlen(iter->name) by calculating it as `iter->name = iter->value + strlen(iter->value) + 1`? this sort of reuses `strlen(name)` from last call

src/kv_list.c
```diff
@@ -0,0 +1,117 @@
...
+
+	/* The list may be NULL or empty. */
+	if (!ns || ns->length < 1)
+		return (kv_iter)
+	{
+		0
+	};
+
+	/* Grab the number of pairs and char pointers to the name and value. */
+	name = (char *) ns + kv_list_data_start;
```
serprex:
```suggestion
	name = ns->data;
```

src/kv_list.c
```diff
@@ -0,0 +1,117 @@
...
+	pairs->length = list_length(list);
+
+	/* in this loop, size reprises the space calculation above */
+	foreach(lc, list)
+	{
+		char	   *str;
+
+		/* Append the element name and arg to the kv_pair. */
+		elem = (DefElem *) lfirst(lc);
+		str = (char *) pairs + ck_size;
```
serprex: have ck_size start as 0, then
```suggestion
		str = pairs->data + ck_size;
```

src/kv_list.c
theory: Oh? How come?

src/kv_list.c
```diff
@@ -0,0 +1,117 @@
...
+
+	/* Count up space for dynamic key/value pairs. */
+	foreach(lc, list)
+	{
+		elem = (DefElem *) lfirst(lc);
+		kv_size += 2 + strlen(elem->defname) + strlen(strVal(elem->arg));
+	}
+
+	/* Alloc the result ... */
+	pairs = allocate_list(kv_size, allocate);
```
serprex: same here, kv_size should start at 0, then `allocate_list` should handle adding the size of the base struct to the kv size param

src/kv_list.c
theory:
```
src/kv_list.c:95:7: warning: assigning to 'char *' from 'const char[]' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
   95 |         name = ns->data;
      |              ^ ~~~~~~~~
```

theory: This works tho:

```c
	name = (char *) ns->data;
```

src/kv_list.c
serprex: seems odd having a long named variable that's only equal to the expression, but I use "I'd prefer" since I recognize most people disagree with the idea of [APL idioms](https://aplwiki.com/wiki/Idiom) preferring to repeat short expressions over introducing variables

src/kv_list.c
theory: Yes, done in b7fd784.

src/kv_list.c
theory: Done in b7fd784.

src/kv_list.c
theory: Yes, done in b7fd784.

src/kv_list.c
theory: Yes, done in b7fd784.

src/kv_list.c
theory: Done in b7fd784.

src/kv_list.c
theory: With the other changes it no longer repeats, so I've removed it in b7fd784.

src/kv_list.c
theory: Huh, the Linux compiler doesn't like it:

```
src/kv_list.c: In function ‘kv_iter_next’:
src/kv_list.c:108:20: warning: operation on ‘iter->name’ may be undefined [-Wsequence-point]
  108 |         iter->name = iter->name = iter->value + strlen(iter->value) + 1;
      |         ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

It's reasonable, since it has no idea that `iter->name` has been set elsewhere. Odd that it doesn't complain about taking its length; maybe because it could be 0?

theory: Oh, never mind, typo, fixed in 709c07f.

serprex (approved):

PR #95, opened 2025-12-13
Add kv_list and refactor session_settings with it

theory:
Compare with #94.

A new structure, kv_list, represents a list of key/value pairs. Its constructor takes a Postgres `List` argument, iterates over the keys and values, allocates memory for each, and copies them into the struct. It uses `malloc()` to allocate the memory, because the we store the value in a global that mustn't be subject to the GUC memory context.

Use it in `chfdw_settings_assign_hook()` to create a `kv_list` from the `List` returned by `chfdw_parse_options()` and assign it to the `ch_session_settings_list` global, which can then be fetched via `chfdw_get_session_settings()`.

The downside to this approach is that the new value must be parsed twice. We consider this acceptable, since it means they're no longer parsed for every query.

While at it, change `chfdw_parse_options()` to use `pstrdup()` instead of `strdup`, since its memory is handled by a Postgres memory context (and surely was leaking before).

### DIFF
--- a/src/kv_list.c
+++ b/src/kv_list.c
@@ -1,0 +1,88 @@
+#include <stddef.h>
+#include "kv_list.h"
+#include "utils/elog.h"
+#include "nodes/pathnodes.h"
+#include "utils/guc.h"
+
+extern kv_list * new_empty_kv_list(void)
+{
+	kv_list    *pairs = (kv_list *) malloc(sizeof(kv_list));
+
+	if (pairs == NULL)
+	{
+		ereport(LOG, errcode(ERRCODE_FDW_OUT_OF_MEMORY), errmsg("out of memory"));
+		return NULL;
+	}
+	pairs->length = 0;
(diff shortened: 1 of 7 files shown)


src/kv_list.c
```diff
@@ -0,0 +1,76 @@
...
+		kv_pair    *pair = malloc(sizeof(kv_pair));
+
+		if (!pair)
+		{
+			kv_list_free(pairs);
+			ereport(LOG, errcode(ERRCODE_FDW_OUT_OF_MEMORY), errmsg("out of memory"));
+			return NULL;
+		}
+		elem = (DefElem *) lfirst(lc);
+		pair->name = strdup(elem->defname);
```
serprex, line 57: need to check for `NULL` from strdup

theory: Yeah, also need to free `name` and `value` in `kv_list_free()`. Done in 676bc13.

PR #109, opened 2025-12-22
Add support for parameterized execution

theory:
The `PREPARE` command requires parameterized execution, as do some SQL queries. For example, PostgreSQL converts this query:

    SELECT * FROM foo
     WHERE id = (SELECT 2);

To this parameterized query:

    SELECT * FROM foo
     WHERE id = $1

postgres_fdw has supported parameterized execution for a long time, and there are vestiges of that feature from the original clickhouse_fdw code included here. Perhaps ClickHouse didn't support parameterized queries back then, but it does now.

Add the functions necessary to build on the existing `params_list`, `numParams`, and `param_values` values to support parameterized execution. Use the ClickHouse parameter format, `{param_name:data_type}`, for parameters in queries, and pass the values via both the binary and http engines.

When using `PREPARE`, the first five executions will send the literal values rather than the parameter placeholders. The sixth and subsequent calls will have the parameter placeholders and pass the parameters separately. Postgres itself decides on this division between a custom plan (with literals) or a generic plan (with placeholders) as described in [PREPARE Notes].

Most of this code comes right from postgres_fdw, mainly to handle cases of `T_Param` nodes. IN addition:

*   Pass the number of params and the params to `new_query()` to be included in a `ch_query`, so that the engines can send them
*   Use `Query.SetParams()` to send params with the binary engine
*   Use a `POST` request with form data to send the query and params with the http engine
*   `process_query_params()` currently converts parameters to text using PostgreSQL's output functions; will need to be updated to consistently use ClickHouse formats (#110)
*   Add the `param.sql` tests ported from postgres_fdw tests

  [PREPARE Notes]: https://www.postgresql.org/docs/current/sql-prepare.html#SQL-PREPARE-NOTES

### DIFF
--- a/src/binary.cpp
+++ b/src/binary.cpp
@@ -184,6 +184,20 @@ static QuerySettings ch_binary_settings(const ch_query *query)
 	return res;
 }
 
+/*
+ * Converts query->param_values to QueryParams.
+ */
+static QueryParams ch_binary_params(const ch_query *query)
+{
+	int i;
+	auto res = QueryParams{};
+
+	for (i = 0; i < query->num_params; i++)
+		res.insert_or_assign(psprintf("p%d", i+1), QueryParamValue(query->param_values[i]));
+
+	return res;
+}
+
 static void set_state_error(ch_binary_read_state_t * state, const char * str)
 {
 	assert(state->error == NULL);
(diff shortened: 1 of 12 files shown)


src/binary.cpp
```diff
@@ -184,6 +184,20 @@ static QuerySettings ch_binary_settings(const ch_query *query)
...
 
+/*
+ * Converts query->param_values to QueryParams.
+ */
+static QueryParams ch_binary_params(const ch_query *query)
+{
+	int i;
+	auto res = QueryParams{};
+
+	for (i = 0; i < query->num_params; i++)
```
serprex, line 195:
```suggestion
	auto res = QueryParams{};

	for (int i = 0; i < query->num_params; i++)
```

theory: Most of the code here declares all variables at the top of each function or block, as Postgres itself does. I'm trying to stick to that pattern (though I prefer declaring stuff as needed).

serprex: I see, didn't expect that pattern to hold in cpp portion of code

PR #113, opened 2025-12-29
Proper param conversion to CH string literals

theory:
The code for passing parameters, originally copied from postgres_fdw, relied upon PostgreSQL's output functions to format string literals. Unfortunately this doesn't work for some ClickHouse string literals, including arrays.

Since the http engine already requires proper string literal generation, extract its code that does so from `extend_insert_query()` into the new, exported `chfdw_datum_to_ch_literal()` function. Simply pass a Datum and its Oid and it returns the appropriate `palloc'd string literal.

However, even that code did not properly handle arrays. `deparse.c` has a function that does it, `deparseArray()`. Export a wrapper around that function, `chfdw_array_to_ch_literal()`, that takes a single `Datum` and constructs the context required for `deparseArray()` to do its job. Then teach `chfdw_datum_to_ch_literal()` to call it for array `Datum`s.

Then replaces the `param_flinfo` variables in `fdw.c` with `param_oids` and store the OIDs for each parameter, rather than a pointer to a Postgres output function. Then change `process_query_params()` to use these OIDs to call `chfdw_datum_to_ch_literal()` to create the parameter string literals.

Add a new test script, `http_inserts.sql`, which duplicates `binary_inserts.sql` but for the http engine. It includes insertion of an array, testing the new functionality.

Also modify the `params.sql` test to convert one of its columns to an array and execute prepared statements with array parameters.

While at it:

*   Rename the mis-spelled `ch_binary_do_output_convertion`

### DIFF
--- a/src/pglink.c
+++ b/src/pglink.c
@@ -340,6 +340,75 @@ chfdw_http_fetch_raw_data(ch_cursor * cursor)
 }
 
 /*
+ * Convert a Datum to a ClickHouse literal string. Returns NULL if the value
+ * cannot be converted to a literal.
+ */
+extern char *
+chfdw_datum_to_ch_literal(Datum value, Oid type)
+{
+	if (type_is_array(type))
+		return chfdw_array_to_ch_literal(value);
+
+	switch (type)
+	{
+		case BOOLOID:
+		case INT2OID:
+		case INT4OID:
+			return psprintf("%d", DatumGetInt32(value));
+		case INT8OID:
+			return psprintf(INT64_FORMAT, DatumGetInt64(value));
+		case FLOAT4OID:
+			return psprintf("%f", DatumGetFloat4(value));
+		case FLOAT8OID:
+			return psprintf("%f", DatumGetFloat8(value));
+		case NUMERICOID:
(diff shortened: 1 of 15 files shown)


src/pglink.c
```diff
@@ -339,6 +339,75 @@ chfdw_http_fetch_raw_data(ch_cursor * cursor)
...
+						   *result;
+				size_t		len;
+				bool		tl = false;
+				Oid			typoutput = InvalidOid;
+
+				getTypeOutputInfo(type, &typoutput, &tl);
+				text = OidOutputFunctionCall(typoutput, value);
+				len = strlen(text) + 1;
+				result = palloc(len * 2 + 1);
+				ch_escape_string(result, text, len);
```
serprex:
```suggestion
				len = strlen(text);
				result = palloc(len * 2 + 1);
				ch_escape_string(result, text, len + 1);
```
seems like this would be a slight overestimate since `+ 1` on `len` shouldn't be part of the `* 2` for `result`

theory: Hah! Fair. Pushed in 96ccdc3.

PR #138, opened 2026-01-26
Map `md5()` to `lower(hex(MD5()))`

theory:
The Postgres `md5()` function returns a hex string, so can be mapped to ClickHouse `lower(hex(MD5()))` pretty easily. Support for other hashing functions will have to wait for proper BYTEA support.

While at it, restore support for passing `TESTS` to `make installcheck` and update the Copyright year.

### DIFF
--- a/test/sql/functions.sql
+++ b/test/sql/functions.sql
@@ -263,7 +263,6 @@ SELECT ts FROM t5 WHERE EXTRACT(month FROM ts::date) = 11;
 EXPLAIN (VERBOSE, COSTS OFF) SELECT ts FROM t5 WHERE EXTRACT(day FROM ts::date) = 18;
 SELECT ts FROM t5 WHERE EXTRACT(day FROM ts::date) = 18;
 
-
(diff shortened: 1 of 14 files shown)


test/sql/functions.sql
```diff
@@ -286,6 +285,10 @@ SELECT ts FROM t5 WHERE date_trunc('quarter', ts) = '2027-10-01'::date;
 EXPLAIN (VERBOSE, COSTS OFF) SELECT val FROM t4 WHERE regexp_like('^val\d', val);
 SELECT val FROM t4 WHERE regexp_like('^val\d', val);
 
+-- Check hashing functions.
+EXPLAIN (VERBOSE, COSTS OFF) SELECT key1, val FROM t3_map WHERE md5(val) LIKE 'a%';
+SELECT key1 FROM t3_map WHERE md5(val) LIKE 'a%';
+
```
serprex: `ORDER BY` would make output stable

PR #141, opened 2026-01-30
Refactor http_fetch_row to resemble binary_fetch_row

theory:
## Refactor fetch_row arg into ChFdwScanRowContext

Rather than pass a bunch of individual arguments to the `fetch_row` function, create a new struct to pass, instead, `ChFdwScanRowContext`. This will simplify adding attributes in the future without editing a bunch of function signatures. Add one new element, the `AttInMetadata` that describes the Postgres input functions for the fields in the row, for use in the next commit.

## Move http datum processing to `http_fetch_row`

Remove the special case handling of rows fetched from the http driver from `fetch_tuple()` in `fdw.c` to `http_fetch_row` in `pglinkc.c`, where it's better encapsulated with the rest of the http engine row parsing. It also better aligns with the behavior of `binary_fetch_row()`.

The logic lives in a new function, `char_to_datum()`, so-named in anticipation of it converting character data to `Datum`s. Right now it simply fixes various issues in the bytes fetched from the HTTP response, same as it previously did in `fetch_tuple()`.

However, also add a second argument, the length of the data being processed, now set by `ch_http_read_next()` in `parser.c`. Not yet used, but will be in the future to support `BYTEA` conversion.

After the text has been properly munged by `char_to_datum()`, iterate through the list of attributes, convert them to `Datum`s, and add them to `ctx->values`. The FDW functionality requires this behavior, while the returning of strings is mainly used for `IMPORT FOREIGN SCHEMA`, which has no `TupleDesc` and works exclusively with strings.

## Teach fetch_row to return Datums instead of void*

Previously, `http_fetch_row()` returned an array of `char *` values while `binary_fetch_row()` returned an array of `Datum`s. Refactor `http_fetch_row()` to also return an array of `Datum`s, and change the `fetch_row` signature accordingly.

To make it work, refactor `http_fetch_row()` to explicitly handle the special case of `SELECT NULL`, just as `binary_fetch_row()` does. Then save on a second iteration through and copying of all the values when `ctx->tupdesc` is set by simply returning the values it scans into `ctx->values`. Update `char_to_datum()` to just insert into `ctx->values` and `ctx->nulls` rather than return a value.

Document these behaviors for both `http_fetch_row()` and binary_fetch_row()`, as it took me quite a while to work out how it all works.

For when `ctx->tupdesc` is not set, process all of the values into either `NULL`s or text `Datum`s. This is the return value used by `chfdw_construct_create_tables()` to create tables, which no no longer must treat http and binary values differently, but treat them all as text `Datum`s.

### DIFF
--- a/src/pglink.c
+++ b/src/pglink.c
@@ -30,10 +30,10 @@ static void http_disconnect(void *conn);
 static ch_cursor * http_simple_query(void *conn, const ch_query * query);
 static void http_simple_insert(void *conn, const ch_query * query);
 static void http_cursor_free(void *);
-static void **http_fetch_row(ChFdwScanRowContext * ctx);
+static Datum * http_fetch_row(ChFdwScanRowContext * ctx);
 static void *http_prepare_insert(void *, ResultRelInfo *, List *, const ch_query *, char *);
 static void http_insert_tuple(void *, TupleTableSlot *);
-static char *char_to_datum(ChFdwScanRowContext * ctx, int attnum, char *data, size_t len);
+static void char_to_datum(ChFdwScanRowContext * ctx, int attnum, char *data, size_t len);
 
 static libclickhouse_methods http_methods =
 {
@@ -49,7 +49,7 @@ static ch_cursor * binary_simple_query(void *conn, const ch_query * query);
 static void binary_cursor_free(void *cursor);
 
 /* static void binary_simple_insert(void *conn, const char *query); */
-static void **binary_fetch_row(ChFdwScanRowContext * ctx);
+static Datum * binary_fetch_row(ChFdwScanRowContext * ctx);
 static void binary_insert_tuple(void *, TupleTableSlot * slot);
 static void *binary_prepare_insert(void *, ResultRelInfo *, List *,
 								   const ch_query * query, char *table_name);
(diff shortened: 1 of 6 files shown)


src/pglink.c
```diff
@@ -325,7 +363,63 @@ http_fetch_row(ch_cursor * cursor, List * attrs, TupleDesc tupdesc, Datum * v, b
...
+			pos[0] = '\0';
+	}
+	else if (data && (pgtype == TIMEOID || pgtype == TIMETZOID)
+			 && data[strlen(data) - 1] == 'Z')
+
+		/*
+		 * date_time_output_format=iso formats times as ISO timestamps. Remove
+		 * the leading `YYYY-mm-ddT`.
+		 */
+		data += strlen("1970-01-01T");
```
serprex, line 424: should wrap this in curlies for legibility

src/pglink.c
```diff
@@ -325,7 +363,63 @@ http_fetch_row(ch_cursor * cursor, List * attrs, TupleDesc tupdesc, Datum * v, b
...
+	/* Easy check array for, else must use get_element_type on pgtype. */
+	if (data[0] == '\\' && data[1] == 'N')
+		data = NULL;
+	else if (data && is_array && data[0] == '[')
+	{
+		size_t		pos = 0;
+
+		while (data[pos] != '\0')
+		{
+			if (data[pos] == '[')
```
serprex, line 399: this seems like it'd mess up with `['[]']` from ClickHouse. Need to add some basic parsing to skip quoted sections

src/pglink.c
theory: Good call, done in 27b6e22. Honestly I'd like to always require braces; omitting them leads to unintentional bugs.

src/pglink.c
theory: Quoted sections? What does that look like? I added a test for an empty array in 27b6e22.

theory: Oh I see what you mean.

theory: Good grief, this array parsing is completely useless for text arrays. I'm going to create a new branch to fix it; as it is, this is just the existing code moved to a new spot. Let's leave it as-is for the PR, if you don't mind.

PR #143, opened 2026-02-03
Use StringInfoData for TSV, drop breaking settings

theory:
Replace the custom string buffer management in `ch_http_read_state` with the Postgres `StringInfo` object, and let it handle memory management and buffer growth as appropriate. This simplifies the code somewhat, and lays the ground for forthcoming TSV parsing improvements.

While at it, explicitly set the `format_tsv_null_representation`, and `output_format_tsv_crlf_end_of_line` settings for all http requests, as any other values will break its TSV parsing.

Also fix the documentation of the default value of the `pg_clickhouse.session_settings` GUC.

### DIFF
--- a/src/http.c
+++ b/src/http.c
@@ -181,13 +181,22 @@ ch_http_simple_query(ch_http_connection_t * conn, const ch_query * query)
 	/* Append each of the settings as a query param. */
 	for (iter = new_kv_iter(query->settings); !kv_iter_done(&iter); kv_iter_next(&iter))
 	{
+		/* Skip settings that would break parsing and type conversion. */
+		if (
+			strcmp(iter.name, "date_time_output_format") == 0 ||
+			strcmp(iter.name, "format_tsv_null_representation") == 0 ||
+			strcmp(iter.name, "output_format_tsv_crlf_end_of_line") == 0
+			)
+			continue;
 		buf = psprintf("%s=%s", iter.name, iter.value);
 		curl_url_set(cu, CURLUPART_QUERY, buf, CURLU_APPENDQUERY | CURLU_URLENCODE);
 		pfree(buf);
 	}
 
-	/* Always use ISO date format */
+	/* Always use ISO date format, \N for NULL, \n for EOL. */
 	curl_url_set(cu, CURLUPART_QUERY, "date_time_output_format=iso", CURLU_APPENDQUERY | CURLU_URLENCODE);
+	curl_url_set(cu, CURLUPART_QUERY, "format_tsv_null_representation=\\N", CURLU_APPENDQUERY | CURLU_URLENCODE);
+	curl_url_set(cu, CURLUPART_QUERY, "output_format_tsv_crlf_end_of_line=0", CURLU_APPENDQUERY | CURLU_URLENCODE);
 	curl_url_get(cu, CURLUPART_URL, &url, 0);
 	curl_url_cleanup(cu);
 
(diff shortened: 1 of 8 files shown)


src/http.c
```diff
@@ -181,6 +181,13 @@ ch_http_simple_query(ch_http_connection_t * conn, const ch_query * query)
 	/* Append each of the settings as a query param. */
 	for (iter = new_kv_iter(query->settings); !kv_iter_done(&iter); kv_iter_next(&iter))
 	{
+		/* Skip settings that would break parsing and type conversion. */
+		if (
```
serprex, line 185: these settings might be here to explicitly fix settings (imagine database has altered these globally)

unless we specify these ourselves explicitly, we should allow these to pass through (or check that their trying to set to value we prefer)

src/parser.c
```diff
@@ -13,86 +13,71 @@ ch_http_read_state_init(ch_http_read_state * state, char *data, size_t datalen)
...
-	state->val = malloc(state->buflen);
+	state->val = makeStringInfo();
 	state->done = false;
 }
 
 void
 ch_http_read_state_free(ch_http_read_state * state)
 {
-	free(state->val);
+	pfree(state->val);
```
serprex, line 25: also need to free `state->val->data`

personally I'd try make `val` StringInfoData instead of StringInfo, removing a layer of indirection

theory: Done in e0a34f7, thanks.

src/include/http.h
```diff
@@ -30,8 +30,7 @@ typedef struct
 	size_t		buflen;
 	size_t		curpos;
 	size_t		maxpos;
-	size_t		len;
-	char	   *val;
+	StringInfoData *val;
```
serprex:
```suggestion
	StringInfoData val;
```

theory: That does not work.

src/http.c
theory: Excellent point, fixed in407bb34.

src/include/http.h
theory: FIWW, If I understand the C correctly, `StringInfo` is just [a pointer](https://github.com/postgres/postgres/blob/57bff90160fdee56a0d55d7eaa7ec5ad709fda08/src/include/lib/stringinfo.h#L54) to a `StringInfoData`:

```
def StringInfoData *StringInfo
```

So it doesn't seem indirect to me to use `StringInfo`.

serprex: Yes, StringInfo is fine without StringInfoData

Using StringInfoData without pointer indirection works, you just have to then pass `&val` around to functions taking StringInfo & initialize StringInfoData with `initStringInfo(&val)`

Completely valid to prefer StringInfo here, but StringInfoData works if you don't intend `val` to outlive state

theory: Seems silly to have to pass it as a reference when it needs to be `palloc`d anyway. I guess it doesn't if it's a StringInfoData, only its `Data` part does.

serprex: yes exactly, this all would just avoid extra 32 byte allocation

theory: I need a primer on C memory management TBH. I presume for the pointer, that otherwise isn't needed, because `&val` just passes the pointer to the address it uses.

serprex: Right, StringInfoData &val passes pointer to space in struct, while StringInfo val is pointer in struct that points to allocation. Then StringInfoData contains another pointer to data, along with length/capacity

Looking at postgres code you'll often find StringInfoData used for local variables when we only want to return the data at end

PR #153, opened 2026-02-24
Fix parsing ClickHouse string with \ at end

serprex:


theory (approved):
Can you add a test case?

theory (approved):

PR #154, opened 2026-02-24
Detect allocation failure & abort curl write in that circumstance

serprex:


src/http.c
serprex, line 46: if realsize is 0 is it correct to just `return 0` at start?

theory: I don't know; [the examples](https://curl.se/libcurl/c/CURLOPT_WRITEFUNCTION.html) don't shed much light on it.

src/http.c
```diff
@@ -47,6 +47,9 @@ static size_t write_data(void *contents, size_t size, size_t nmemb, void *userp)
 	else
 		res->data = realloc(res->data, res->datasize + realsize + 1);
 
+	if (res->data == NULL)
+		return CURL_WRITEFUNC_ERROR;
+
```
theory, line 52: Has this come up? Possible to make a test case? Seems like it'd be tricky…

src/http.c
serprex: don't expect 0 to happen naturally, so should be fine to skip optimizing for it, code as is works fine with 0

theory (approved):

src/http.c
serprex: would need fault injection on allocator to write reliable test

PR #164, opened 2026-04-01
Fix ILIKE and POSIX regex operator pushdown

iskakaushik:
## Summary

- **ILIKE** (`~~*`) was deparsed as case-sensitive `LIKE`, silently
  returning wrong results. Now correctly emits `ILIKE` / `NOT ILIKE`,
  which ClickHouse supports natively.
- **`~` / `!~`** (POSIX regex) were pushed verbatim, causing syntax
  errors. Now rewritten to `match(col, pattern)` /
  `NOT match(col, pattern)`.
- **`~*` / `!~*`** (case-insensitive regex) are marked non-shippable
  and fall back to local PostgreSQL evaluation, since ClickHouse has
  no direct equivalent.

## Test plan

- [x] New regression test `ilike_regex` covers all operator variants
  (ILIKE, NOT ILIKE, LIKE, NOT LIKE, `~`, `!~`, `~*`, `!~*`) with
  both EXPLAIN plan verification and result correctness checks
- [x] All 21 regression tests pass (20 existing + 1 new)

### DIFF
--- a/src/custom_types.c
+++ b/src/custom_types.c
@@ -376,8 +376,10 @@ classify_builtin_operator(const char *oprname)
 		return CF_REGEX_MATCH;
 	if (strcmp(oprname, "!~") == 0)
 		return CF_REGEX_NO_MATCH;
-	if (strcmp(oprname, "~*") == 0 || strcmp(oprname, "!~*") == 0)
-		return CF_UNSHIPPABLE;
+	if (strcmp(oprname, "~*") == 0)
+		return CF_REGEX_ICASE_MATCH;
+	if (strcmp(oprname, "!~*") == 0)
+		return CF_REGEX_ICASE_NO_MATCH;
 	return CF_USUAL;
 }
 
@@ -406,7 +408,9 @@ chfdw_check_for_custom_operator(Oid opoid, Form_pg_operator form)
 				{
 					tuple = SearchSysCache1(OPEROID, ObjectIdGetDatum(opoid));
 					if (!HeapTupleIsValid(tuple))
-						elog(ERROR, "cache lookup failed for operator %u", opoid);
+						ereport(ERROR,
+								errcode(ERRCODE_INTERNAL_ERROR),
+								errmsg("pg_clickhouse: cache lookup failed for operator %u", opoid));
 					form = (Form_pg_operator) GETSTRUCT(tuple);
 				}
 
--- a/src/custom_types.c
+++ b/src/custom_types.c
@@ -362,10 +362,30 @@ chfdw_check_for_custom_type(Oid typeoid)
 	return entry;
 }
 
+/*
+ * Map a builtin operator name to its custom_object_type.  Returns CF_USUAL
+ * when the operator needs no special handling and should follow the normal
+ * builtin shortcut (i.e. be presumed shippable with no rewrite).
+ *
+ * Keep this function in sync with the CF_* enum in fdw.h.
+ */
+static custom_object_type
+classify_builtin_operator(const char *oprname)
+{
+	if (strcmp(oprname, "~") == 0)
+		return CF_REGEX_MATCH;
+	if (strcmp(oprname, "!~") == 0)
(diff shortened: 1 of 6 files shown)


src/include/fdw.h
```diff
@@ -296,6 +296,9 @@ typedef enum
 	CF_INTARRAY_IDX,
 	CF_CH_FUNCTION,				/* adapted clickhouse function */
 	CF_MATCH,					/* regexp_match function */
+	CF_MD5,						/* md5() function */
```
theory: `CF_MD5` was removed in 9d7e0cc so you don't need it here.

src/custom_types.c
```diff
@@ -380,6 +400,22 @@ chfdw_check_for_custom_operator(Oid opoid, Form_pg_operator form)
 			case F_TIMESTAMPTZ_PL_INTERVAL:
 				break;
 			default:
+
+				/* Look up the operator name so we can classify it. */
+				if (!form)
+				{
+					tuple = SearchSysCache1(OPEROID, ObjectIdGetDatum(opoid));
+					if (!HeapTupleIsValid(tuple))
+						elog(ERROR, "cache lookup failed for operator %u", opoid);
```
theory: Let's try to use `ereport` with an appropriate error code. I'd like to get all the errors updated to use it (#167), but for now I think it's enough for new errors to use it.

src/custom_types.c
```diff
@@ -362,10 +362,30 @@ chfdw_check_for_custom_type(Oid typeoid)
...
+ *
+ * Keep this function in sync with the CF_* enum in fdw.h.
+ */
+static custom_object_type
+classify_builtin_operator(const char *oprname)
+{
+	if (strcmp(oprname, "~") == 0)
+		return CF_REGEX_MATCH;
+	if (strcmp(oprname, "!~") == 0)
+		return CF_REGEX_NO_MATCH;
```
theory, line 378: This is fine for now, but long term I expect we're going to want a data structure that maps operator names, somewhat as we currently do for functions.

test/expected/ilike_regex.out
```diff
@@ -0,0 +1,305 @@
+-- Test ILIKE and POSIX regex operator pushdown
+SET datestyle = 'ISO';
+CREATE SERVER ilike_regex_loopback FOREIGN DATA WRAPPER clickhouse_fdw
+    OPTIONS(dbname 'ilike_regex_test', driver 'binary');
```
theory, line 4: We should also test with the http driver.

test/expected/ilike_regex.out
```diff
@@ -0,0 +1,305 @@
...
+   Output: (count(*))
+   Relations: Aggregate on (events)
+   Remote SQL: SELECT count(*) FROM ilike_regex_test.events WHERE ((name ILIKE 'page______'))
+(4 rows)
+
+SELECT count(*) FROM events WHERE name ILIKE 'page______';
+ count 
+-------
+     0
+(1 row)
```
theory: We should have a variant of this test that returns a matching result matched by the underscore wildcard.

src/custom_types.c
```diff
@@ -362,10 +362,30 @@ chfdw_check_for_custom_type(Oid typeoid)
...
+	if (strcmp(oprname, "~*") == 0 || strcmp(oprname, "!~*") == 0)
+		return CF_UNSHIPPABLE;
```
theory: I'm surprised there's no case-insensitive variant of [match](https://clickhouse.com/docs/sql-reference/functions/string-search-functions#match)!

theory (changes requested):

src/custom_types.c
theory: We could prepend `(?i)` to the regex to achieve this. [Example](https://fiddle.clickhouse.com/48cba713-0852-4fc1-b73c-7422f4c10064).

iskakaushik: great idea!!

theory (approved):

PR #169, opened 2026-04-02
Push down jsonb -> and ->> operators as ClickHouse dot notation

iskakaushik:
## Summary

- Rewrite jsonb `->` and `->>` operators to ClickHouse dot notation
  during deparse (e.g. `data ->> 'name'` becomes `data.name`), enabling
  ClickHouse to filter on JSON fields instead of shipping entire rows
  for local evaluation.
- Keys are emitted through `quote_identifier`, correctly quoting spaces,
  dots, mixed case, reserved words, and embedded double quotes.
- Follows up on bbdb1a6 which disabled subscript (`column['key']`)
  pushdown; this provides the working alternative via `->>`/`->`.

## Test plan

- [x] `EXPLAIN` verifies WHERE equality pushdown: `data.name = 'widget'`
- [x] `EXPLAIN` verifies WHERE LIKE pushdown: `data.name LIKE 'wid%'`
- [x] `EXPLAIN` verifies compound WHERE (AND / OR) pushdown
- [x] `EXPLAIN` verifies ORDER BY pushdown: `ORDER BY data.size`
- [x] `EXPLAIN` verifies `->` operator pushdown (returns jsonb, not text)
- [x] Edge cases: keys with spaces, mixed case, reserved words, embedded
  double quotes, backslashes, dots, apostrophes, special characters, and
  leading digits -- all produce correctly quoted remote SQL
- [x] Data fetches return correct values for all edge cases
- [x] All 4 expected output variants updated (PG 14-18, PG 13, CH 24.8-25.3, CH 22-24.3)
- [x] `make tempcheck` passes (json test green; pre-existing failures in
  functions, http, http_inserts, import_schema are unrelated)

### DIFF
--- a/src/custom_types.c
+++ b/src/custom_types.c
@@ -363,23 +363,37 @@ chfdw_check_for_custom_type(Oid typeoid)
 }
 
 /*
+ * Operator-name to custom_object_type mapping table, searched linearly by
+ * classify_builtin_operator().  Keep in sync with the CF_* enum in fdw.h.
+ */
+typedef struct
+{
+	const char *oprname;
+	custom_object_type ctype;
+}			BuiltinOperatorMap;
+
+static const BuiltinOperatorMap builtin_operator_map[] = {
+	{"~", CF_REGEX_MATCH},
+	{"!~", CF_REGEX_NO_MATCH},
+	{"~*", CF_REGEX_ICASE_MATCH},
+	{"!~*", CF_REGEX_ICASE_NO_MATCH},
+	{"->", CF_JSONB_FETCHVAL},
+	{"->>", CF_JSONB_FETCHVAL_TEXT},
+};
+
+/*
  * Map a builtin operator name to its custom_object_type.  Returns CF_USUAL
  * when the operator needs no special handling and should follow the normal
  * builtin shortcut (i.e. be presumed shippable with no rewrite).
- *
- * Keep this function in sync with the CF_* enum in fdw.h.
  */
 static custom_object_type
 classify_builtin_operator(const char *oprname)
 {
-	if (strcmp(oprname, "~") == 0)
-		return CF_REGEX_MATCH;
-	if (strcmp(oprname, "!~") == 0)
-		return CF_REGEX_NO_MATCH;
-	if (strcmp(oprname, "~*") == 0)
-		return CF_REGEX_ICASE_MATCH;
-	if (strcmp(oprname, "!~*") == 0)
-		return CF_REGEX_ICASE_NO_MATCH;
+	for (int i = 0; i < lengthof(builtin_operator_map); i++)
+	{
+		if (strcmp(oprname, builtin_operator_map[i].oprname) == 0)
+			return builtin_operator_map[i].ctype;
+	}
 	return CF_USUAL;
 }
 
--- a/src/deparse.c
+++ b/src/deparse.c
@@ -2745,6 +2745,45 @@ deparseOpExpr(OpExpr * node, deparse_expr_cxt * context)
 					goto cleanup;
 				}
 				break;
+			case CF_JSONB_FETCHVAL:
+			case CF_JSONB_FETCHVAL_TEXT:
+				{
+					Expr	   *arg1 = linitial(node->args);
+					Expr	   *arg2 = lsecond(node->args);
+
+					/*
+					 * Convert jsonb ->> 'key' to ClickHouse dot notation:
+					 * column.key Convert jsonb -> 'key' to JSON-wrapped dot
+					 * notation: toJSONString(column.key) The -> operator
+					 * returns jsonb, so we need to wrap the result with
+					 * toJSONString() for type compatibility.
+					 */
+					if (IsA(arg2, Const))
+					{
+						Const	   *key = (Const *) arg2;
+
+						if (key->consttype == TEXTOID && !key->constisnull)
+						{
+							char	   *keyval = TextDatumGetCString(key->constvalue);
+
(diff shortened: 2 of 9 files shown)


src/custom_types.c
```diff
@@ -380,6 +380,10 @@ classify_builtin_operator(const char *oprname)
 		return CF_REGEX_ICASE_MATCH;
 	if (strcmp(oprname, "!~*") == 0)
 		return CF_REGEX_ICASE_NO_MATCH;
+	if (strcmp(oprname, "->") == 0)
+		return CF_JSONB_FETCHVAL;
+	if (strcmp(oprname, "->>") == 0)
+		return CF_JSONB_FETCHVAL_TEXT;
```
theory: Can we switch from these special constants to a table that maps operators? Something similar to `chfdw_check_for_custom_function`'s handling of functions.

src/deparse.c
```diff
@@ -2745,6 +2745,35 @@ deparseOpExpr(OpExpr * node, deparse_expr_cxt * context)
 					goto cleanup;
 				}
 				break;
+			case CF_JSONB_FETCHVAL:
+			case CF_JSONB_FETCHVAL_TEXT:
```
theory, line 2749: These are different; I think `CF_JSONB_FETCHVAL` needs to convert results to JSON strings for compatibility.

test/expected/json.out
```diff
@@ -136,6 +136,246 @@ SELECT data['size'], count(*) FROM json_http.things GROUP BY data['size'];
...
 
+-- The jsonb ->> operator should be pushed down to ClickHouse using dot
+-- notation (e.g. data ->> 'name' becomes data.name on the remote side).
+EXPLAIN (VERBOSE, COSTS OFF)
+SELECT data ->> 'name' FROM json_http.things;
+                   QUERY PLAN                    
+-------------------------------------------------
+ Foreign Scan on json_http.things
+   Output: (data ->> 'name'::text)
+   Remote SQL: SELECT data FROM json_test.things
```
theory, line 149: Comment suggests this should be pushed down, but it's not (and won't be). Suggest removing it or commenting that this is expected.

test/expected/json.out
```diff
@@ -136,6 +136,246 @@ SELECT data['size'], count(*) FROM json_http.things GROUP BY data['size'];
...
+(3 rows)
+
+-- The jsonb -> operator should also be pushed down using dot notation.
+EXPLAIN (VERBOSE, COSTS OFF)
+SELECT data -> 'name' FROM json_http.things;
+                   QUERY PLAN                    
+-------------------------------------------------
+ Foreign Scan on json_http.things
+   Output: (data -> 'name'::text)
+   Remote SQL: SELECT data FROM json_test.things
```
theory, line 227: Same comment as above that this lack of pushdown is expected.

theory: But also, there is no example here of `->` pushdown. If there were I suspect it would not work correctly, because `->` expects a JSON value, whereas the [dot notation](https://clickhouse.com/docs/sql-reference/data-types/newjson#reading-json-paths-as-sub-columns) seems to return strings. That means that something like this wouldn't find a record:

```sql
SELECT * FROM json_http.things WHERE data -> 'name' = '"widget"'::json;
```

test/expected/json.out
```diff
@@ -136,6 +136,246 @@ SELECT data['size'], count(*) FROM json_http.things GROUP BY data['size'];
...
+(2 rows)
+
+-- Key with mixed case.
+EXPLAIN (VERBOSE, COSTS OFF)
+SELECT * FROM json_http.special_keys WHERE data ->> 'CamelCase' = 'world';
+                                           QUERY PLAN                                           
+------------------------------------------------------------------------------------------------
+ Foreign Scan on json_http.special_keys
+   Output: id, data
+   Remote SQL: SELECT id, data FROM json_test.special_keys WHERE ((data."CamelCase" = 'world'))
```
theory, line 325: Nice.

theory (changes requested):
This is great. `->`, however, does not work properly. We need to cast its output to JSON. Is that possible? IIRC ClickHouse only supports objects, but surely they can be nested.

test/expected/json.out
```diff
@@ -136,6 +136,259 @@ SELECT data['size'], count(*) FROM json_http.things GROUP BY data['size'];
...
+-- WHERE clause with -> wraps the dot notation in toJSONString() so that the
+-- result is a proper JSON value (-> returns jsonb, not text like ->>).
+EXPLAIN (VERBOSE, COSTS OFF)
+SELECT * FROM json_http.things WHERE data -> 'name' = '"widget"'::jsonb;
+                                             QUERY PLAN                                             
+----------------------------------------------------------------------------------------------------
+ Foreign Scan on json_http.things
+   Output: id, data
+   Remote SQL: SELECT id, data FROM json_test.things WHERE ((toJSONString(data.name) = '"widget"'))
+(3 rows)
```
theory, line 282: Would like to see the actual output from this query, which is not currently run on ClickHouse. Would be useful to have it return a few types — boolean, number, string, object — just to ensure that we get the expected results.

theory (approved):
Looks good, just would like a few more `->` tests to ensure `toJSONString()` does what we think it does.

After that, please rebase and create a single commit that describes the changes. Thank you!

test/expected/json.out
theory: Added in 0b4c03e.

test/expected/json.out
theory: Added in 0b4c03e.

PR #171, opened 2026-04-02
fuzz parser

serprex:
fix bug found in parser.c: if response is MAX_INPUT in size without trailing newline,
then read would get to MAX_INPUT+1's zero byte & read out of bounds

### DIFF
--- a/fuzz/Makefile
+++ b/fuzz/Makefile
@@ -1,0 +1,48 @@
+AFL_CC    ?= afl-clang-fast
+CC        ?= clang
+CFLAGS    ?= -g -O2 -fsanitize=address,undefined -fno-omit-frame-pointer
+SRC_DIR    = ../src
+INCLUDE    = -I. -I$(SRC_DIR)/include
+
+# Stub out curl/curl.h
+CFLAGS    += -include pg_stub.h -Dcurl_off_t=long $(INCLUDE)
+
+# Override postgres.h via pg_stub.h
+# parser.c includes http.h -> postgres.h, nodes/pg_list.h, lib/stringinfo.h, engine.h
+
+SRCS       = fuzz_parser.c $(SRC_DIR)/parser.c
+
+.PHONY: all clean fuzz triage
+
+all: fuzz_parser
+
+STUB_HEADERS = stubs/postgres.h stubs/nodes/pg_list.h stubs/nodes/pathnodes.h \
+               stubs/lib/stringinfo.h stubs/engine.h stubs/curl/curl.h
+
+fuzz_parser: $(SRCS) $(STUB_HEADERS)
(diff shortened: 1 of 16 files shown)


fuzz/stubs/curl/curl.h
theory, line 1: Why are there whitespace changes in a bunch of files here?

fuzz/Makefile
```diff
@@ -0,0 +1,33 @@
...
+
+all: $(HARNESS)
+
+STUB_HEADERS = stubs/postgres.h stubs/nodes/pg_list.h stubs/nodes/pathnodes.h \
+               stubs/lib/stringinfo.h stubs/engine.h stubs/curl/curl.h
+
+$(HARNESS): fuzz_parser.c $(SRC_DIR)/parser.c $(STUB_HEADERS)
+	$(AFL_CC) $(CFLAGS) -I stubs -o $@ fuzz_parser.c $(SRC_DIR)/parser.c
+
+# Stub headers — empty files satisfying #include directives
```
theory, line 31: Oooh. Question answered. Maybe they should have comments in each explaining this purpose.

fuzz/pg_stub.h
```diff
@@ -0,0 +1,104 @@
+/*
+ * Minimal PostgreSQL stubs for fuzzing parser.c standalone.
```
theory:
```suggestion
 * Minimal PostgreSQL and libcurl stubs for fuzzing parser.c standalone.
```

fuzz/README.md
```diff
@@ -0,0 +1,38 @@
...
+values. PostgreSQL dependencies are stubbed so the harness runs standalone.
+
+## Build
+
+```sh
+make -C fuzz
+```
+
+Requires `afl-clang-fast` (AFL++ with LLVM mode). Override `AFL_CC` to use a
+different compiler, e.g. `make -C fuzz AFL_CC=afl-gcc`.
```
theory, line 14: Could use some links to installation instructions. NFI how to install on macOS.

theory (changes requested):
Would you mind giving me a demo? Would like to try it on my Mac. Should we also add a GH workflow to run it on every commit?

fuzz/README.md
serprex: http://aflplus.plus might be where to give info, `pg_bsd_ident` is much more pain on Arch (no packages, outdated AUR, tried to build it myself & hit some error..)

fuzz/Makefile
serprex: I think in practice it's clear having them all in `stubs` directory, does make a funny diff tho

theory:
No I'm happy to add something like this, just needs a little more documentation, IMO.

serprex:
I think it'd be better to have a full end to end fuzzer merged, this was more POC

PR #178, opened 2026-04-03
pglink.c: handle OOM preventing error being set

serprex:
`Assert(ch_error)` is incorrect since it's sourced from `strdup(e.what())`


iskakaushik (approved):

theory (approved):
Yay bug fix!

Can you explain a bit more what this is about in the commit message, and also note the bug fix in `CHANGELOG.md`, please?=

PR #179, opened 2026-04-03
Fix appending Merge suffix incorrectly

serprex:
Regression tests intermittently fail:
```diff
diff -U3 /__w/pg_clickhouse/pg_clickhouse/test/expected/engines_1.out /__w/pg_clickhouse/pg_clickhouse/results/engines.out
--- /__w/pg_clickhouse/pg_clickhouse/test/expected/engines_1.out	2026-04-03 02:54:09.175161303 +0000
+++ /__w/pg_clickhouse/pg_clickhouse/results/engines.out	2026-04-03 02:54:33.620083704 +0000
@@ -247,7 +247,7 @@
  Foreign Scan
    Output: a, ((percentile_cont('0.75'::double precision) WITHIN GROUP (ORDER BY ((f)::double precision)) / (sum(d))::double precision))
    Relations: Aggregate on (t4)
-   Remote SQL: SELECT a, (quantileMerge(0.75)(f) / sum(d)) FROM engines_test.t4 GROUP BY a
+ exit 2
+   Remote SQL: SELECT a, (quantile(0.75)(f) / sum(d)) FROM engines_test.t4 GROUP BY a
 (4 rows)
 
 SELECT a, percentile_cont(0.75) WITHIN GROUP (ORDER BY f) / sum(d) FROM t4 GROUP BY a;
```

The compound expression `percentile_cont(0.75) WITHIN GROUP (ORDER BY f) / sum(d)` deparses to `quantile(0.75)(f)` instead of `quantileMerge(0.75)(f)`

Yet the standalone `percentile_cont(0.75) WITHIN GROUP (ORDER BY f)` in the same test run always deparses correctly with the `Merge` suffix

## old code

`foreign_expr_walker` checked whether expressions are safe to push down. As a side effect, when a `T_Var` inside an `T_Aggref` referenced an `AggregateFunction` column, the walker set `agg->location = -2` on the enclosing Aggref:
```c
/* T_Var handler */
cinfo = chfdw_get_custom_column_info(rte->relid, var->varattno);
if (cinfo && cinfo->is_AggregateFunction == CF_AGGR_FUNC)
    outer_cxt->found_AggregateFunction = true;

/* T_Aggref handler, per-arg loop */
inner_cxt.found_AggregateFunction = false;
foreign_expr_walker(n, glob_cxt, &inner_cxt);
if (inner_cxt.found_AggregateFunction)
    agg->location = -2;
inner_cxt.found_AggregateFunction = false;
```

`deparseAggref` read back the flag:
```c
if (node->location == -2)
    appendStringInfoString(buf, "Merge");
```

### `add_to_flat_tlist` copies nodes

After `foreign_grouping_ok` calls `chfdw_is_foreign_expr` setting `location = -2` on the Aggref,
it then stores the expression in `fpinfo->grouped_tlist` via [`add_to_flat_tlist`][add_to_flat_tlist]

`add_to_flat_tlist` wraps expressions with `makeTargetEntry(copyObject(expr), ...)`. Correct behavior relied on `location = -2` happening before the copy

Any planner operation that copies, normalizes, or reconstructs PathTarget expressions between the walker and the tlist construction would produce Aggref nodes with the original parser `location`

[`make_pathtarget_from_tlist`][make_pathtarget] builds PathTarget by appending `tle->expr` without `copyObject`. PathTarget shares pointers with the query's target list. The query optimizer is free to rebuild these at any point: it may re-flatten target lists, substitute equivalent nodes, or reconstruct PathTargets across planner phases

When such an operation produces a fresh Aggref, the new node carries the original parser `location`, losing `location = -2`

### `location` is excluded from equality comparisons

The `Aggref` struct's [`location` field][aggref_location] has type `ParseLoc`. PostgreSQL's node support generator ([`gen_node_support.pl`][gen_node_support]) emits `COMPARE_LOCATION_FIELD` for `ParseLoc` fields, which is defined as [a no-op][compare_location_noop]

This means two Aggref nodes that differ *only* in `location` are
considered equal. If `add_to_flat_tlist` finds a [`tlist_member`][tlist_member] match,
it discards our mutated copy

### source of flakiness

Node identity depends on the exact sequence of planner operations, which varies with PostgreSQL version, table statistics, expression complexity, and cost estimates. A standalone Aggref passes through fewer planner transformations than a compound expression containing casts and multiple aggregates, which is why the standalone query is stable while the compound one is not

`chfdw_get_custom_column_info` cannot be used because `custom_columns_cache` can be invalidated by time `deparseAggref` is called. See `AcceptInvalidationMessages`

[aggref_location]: https://github.com/postgres/postgres/blob/b7b27eb41a5cc0b45a1a9ce5c1cde5883d7bc358/src/include/nodes/primnodes.h
    "Aggref struct — location field typed ParseLoc"

[add_to_flat_tlist]: https://github.com/postgres/postgres/blob/b7b27eb41a5cc0b45a1a9ce5c1cde5883d7bc358/src/backend/optimizer/util/tlist.c
    "add_to_flat_tlist — uses makeTargetEntry(copyObject(expr), ...)"

[make_pathtarget]: https://github.com/postgres/postgres/blob/b7b27eb41a5cc0b45a1a9ce5c1cde5883d7bc358/src/backend/optimizer/util/tlist.c
    "make_pathtarget_from_tlist — appends tle->expr without copying"

[compare_location_noop]: https://github.com/postgres/postgres/blob/b7b27eb41a5cc0b45a1a9ce5c1cde5883d7bc358/src/backend/nodes/equalfuncs.c#L80
    "COMPARE_LOCATION_FIELD defined as ((void) 0)"

[gen_node_support]: https://github.com/postgres/postgres/blob/b7b27eb41a5cc0b45a1a9ce5c1cde5883d7bc358/src/backend/nodes/gen_node_support.pl
    "Node support generator — ParseLoc triggers COMPARE_LOCATION_FIELD"

[tlist_member]: https://github.com/postgres/postgres/blob/b7b27eb41a5cc0b45a1a9ce5c1cde5883d7bc358/src/backend/optimizer/util/tlist.c#L141
    "tlist_member — uses equal() to find matching expressions"

[pull_var_clause]: https://github.com/postgres/postgres/blob/b7b27eb41a5cc0b45a1a9ce5c1cde5883d7bc358/src/backend/optimizer/util/var.c
    "pull_var_clause — returns original node pointers, no copies"

### DIFF
--- a/src/deparse.c
+++ b/src/deparse.c
@@ -65,12 +65,6 @@ typedef struct foreign_glob_cxt
 								 * scan */
 }			foreign_glob_cxt;
 
-typedef struct foreign_loc_cxt
-{
-	bool		found_AggregateFunction;	/* indicator or CH
-											 * AggregateFunction fields */
-}			foreign_loc_cxt;
-
 /*
  * Context for deparseExpr
  */
@@ -109,8 +103,7 @@ do { \
  * remote server.
  */
 static bool foreign_expr_walker(Node * node,
-								foreign_glob_cxt * glob_cxt,
-								foreign_loc_cxt * outer_cxt);
+								foreign_glob_cxt * glob_cxt);
 static char *deparse_type_name(Oid type_oid, int32 typemod);
 
 /*
@@ -223,7 +216,6 @@ chfdw_is_foreign_expr(PlannerInfo * root,
 					  Expr * expr)
 {
 	foreign_glob_cxt glob_cxt;
-	foreign_loc_cxt loc_cxt = {false};
 	CHFdwRelationInfo *fpinfo = (CHFdwRelationInfo *) (baserel->fdw_private);
 
 	/*
@@ -243,7 +235,7 @@ chfdw_is_foreign_expr(PlannerInfo * root,
 	else
 		glob_cxt.relids = baserel->relids;
 
-	if (!foreign_expr_walker((Node *) expr, &glob_cxt, &loc_cxt))
+	if (!foreign_expr_walker((Node *) expr, &glob_cxt))
 		return false;
 
 	/* OK to evaluate on the remote server */
@@ -276,8 +268,6 @@ chfdw_is_equal_op(Oid opno)
 /*
  * Check if expression is safe to execute remotely, and return true if so.
  *
- * In addition, *outer_cxt is updated with collation information.
- *
  * We must check that the expression contains only node types we can deparse,
  * that all types/functions/operators are safe to send (they are "shippable"),
  * and that all collations used in the expression derive from Vars of the
@@ -288,12 +278,10 @@ chfdw_is_equal_op(Oid opno)
  */
 static bool
 foreign_expr_walker(Node * node,
-					foreign_glob_cxt * glob_cxt,
-					foreign_loc_cxt * outer_cxt)
+					foreign_glob_cxt * glob_cxt)
 {
 	bool		check_type = true;
 	CHFdwRelationInfo *fpinfo;
-	foreign_loc_cxt inner_cxt = {false};
 
 	/* Need do nothing for empty subexpressions */
 	if (node == NULL)
@@ -318,9 +306,6 @@ foreign_expr_walker(Node * node,
 				if (bms_is_member(var->varno, glob_cxt->relids) &&
 					var->varlevelsup == 0)
 				{
-					RangeTblEntry *rte;
-					CustomColumnInfo *cinfo;
-
 					/* Var belongs to foreign table */
 
 					/*
@@ -331,17 +316,6 @@ foreign_expr_walker(Node * node,
 					 */
 					if (var->varattno < 0)
 						return false;
-
-					rte = planner_rt_fetch(var->varno, glob_cxt->root);
-					cinfo = chfdw_get_custom_column_info(rte->relid, var->varattno);
-
-					/*
-					 * If this T_Var column is an aggregate function column,
-					 * tell the outer context as much, so that it propagates
-					 * to its parent `T_Aggref`, if any.
-					 */
-					if (cinfo && cinfo->is_AggregateFunction == CF_AGGR_FUNC)
-						outer_cxt->found_AggregateFunction = true;
 				}
 			}
 			break;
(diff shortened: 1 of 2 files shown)


iskakaushik (approved):

src/deparse.c
```diff
@@ -3224,6 +3179,52 @@ appendAggOrderBy(List * orderList, List * targetList, deparse_expr_cxt * context
 	}
 }
 
+/*
+ * Check if Aggref operates on an AggregateFunction column.
```
theory: Would you expand this comment a bit to explain what an AggregateFunction is, why we care, and *how* it makes that determination?

src/deparse.c
```diff
@@ -583,25 +553,13 @@ foreign_expr_walker(Node * node,
...
-					 * an AggregateFunction. If so, we'll append "Merge" to
-					 * the function call in deparseAggref.
-					 */
-					inner_cxt.found_AggregateFunction = false;
-					if (!foreign_expr_walker(n, glob_cxt, &inner_cxt))
+					if (!foreign_expr_walker(n, glob_cxt))
 						return false;
-
-					if (inner_cxt.found_AggregateFunction)
-						agg->location = -2;
```
theory, line 619: This always looked odd to me, like overloading location for another purpose. Makes sense to remove it.

src/deparse.c
```diff
@@ -287,12 +277,10 @@ chfdw_is_equal_op(Oid opno)
  */
 static bool
 foreign_expr_walker(Node * node,
-					foreign_glob_cxt * glob_cxt,
-					foreign_loc_cxt * outer_cxt)
+					foreign_glob_cxt * glob_cxt)
```
theory, line 281: Just to note, postgres_fdw has:

```c
foreign_expr_walker(Node *node,
					foreign_glob_cxt *glob_cxt,
					foreign_loc_cxt *outer_cxt,
					foreign_loc_cxt *case_arg_cxt)
```

It appears to use `outer_ctx` for collation info (which I don't think we can determine, since ClickHouse doesn't support them, IIRC). For `case_arg_ctx`, the comments say:

```
 * case_arg_cxt is NULL if this subexpression is not inside a CASE-with-arg.
 * Otherwise, it points to the collation info derived from the arg expression,
 * which must be consulted by any CaseTestExpr.
```

I suspect we'll want that at some point, but not essential here.

theory (changes requested):

src/deparse.c
theory: Also please describe the fix in `CHANGELOG.md`. TY!

theory (approved):

PR #126, opened 2026-01-13
SubPlan Pushdown

theory:

PR #186, opened 2026-04-04
SubPlan Pushdown

serprex:
rebased #126 & got working with tests


src/deparse.c
```diff
@@ -1953,9 +1951,12 @@ deparseExpr(Expr * node, deparse_expr_cxt * context)
 					if (sp->subLinkType == ANY_SUBLINK)
 					{
 						deparseSubPlan(sp, context);
+						found = true;
 						break;
 					}
 				}
+				if (!found)
+					elog(ERROR, "no ANY_SUBLINK alternative for pushdown");
```
serprex: should be ereport, would like to get this cleaned up across codebase after http streaming lands

theory (changes requested):
I pushed a commit that executes the fully pushed-down query. It fails. I suspect we need to do some sort of jiggerpokery to generate new aliases in the subquery and then to reference tables in the outer query.

test/expected/subquery_pushdown.out
```diff
@@ -160,6 +160,535 @@ ORDER BY o_orderkey;
...
+ORDER BY id;
+                                                                                 QUERY PLAN                                                                                 
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+ Foreign Scan on subquery_test.spd_orders
+   Output: spd_orders.id, spd_orders.status
+   Remote SQL: SELECT id, status FROM subquery_test.spd_orders WHERE ((id IN (SELECT order_id FROM subquery_test.items) OR (status = 'closed'))) ORDER BY id ASC NULLS LAST
+   SubPlan 2
+     ->  Foreign Scan on subquery_test.items
+           Output: items.order_id
+           Remote SQL: SELECT order_id FROM subquery_test.items
```
theory, line 268: Can this and the other queries in this file not be pushed down as a single ClickHouse query?

src/deparse.c
```diff
@@ -374,6 +388,66 @@ foreign_expr_walker(Node * node,
...
+		case T_SubPlan:
+			{
+				SubPlan    *subplan = (SubPlan *) node;
+				PlannerInfo *subroot;
+				Query	   *subquery;
+				ListCell   *lc;
+
+				if (subplan->subLinkType != ANY_SUBLINK &&
+					subplan->subLinkType != EXISTS_SUBLINK &&
+					subplan->subLinkType != EXPR_SUBLINK)
```
theory, line 432: So, no support for CTEs (yet)? What about the others?

src/deparse.c
```diff
@@ -374,6 +388,66 @@ foreign_expr_walker(Node * node,
...
+
+					if (rte->rtekind != RTE_RELATION)
+						return false;
+
+					ft = GetForeignTable(rte->relid);
+					if (ft->serverid != fpinfo->server->serverid)
+						return false;
+				}
+
+				/* Reject nested SubPlans (correlated or InitPlan) */
```
theory: Why?

src/deparse.c
```diff
@@ -2326,6 +2483,207 @@ deparseParam(Param * node, deparse_expr_cxt * context)
...
+		first = false;
+
+		deparseExpr(tle->expr, &context);
+	}
+
+	/* EXISTS subqueries may have all-resjunk targets */
+	if (first)
+		appendStringInfoChar(buf, '1');
+
+	/* FROM clause */
```
theory, line 2661: Are we unable to use the existing deparsers for these clauses? `deparseFromExprForRel()` or `deparseFromExpr()`, `deparseSortGroupClause()`, etc.?

src/deparse.c
```diff
@@ -380,6 +403,101 @@ foreign_expr_walker(Node * node,
...
+				 * Unsupported (and why):
+				 *   ALL_SUBLINK         CH lacks ALL; would need NOT EXISTS
+				 *                       rewrite. NOT IN already arrives as
+				 *                       NOT(ANY_SUBLINK).
+				 *   ROWCOMPARE_SUBLINK  Multi-column compare not deparsed.
+				 *   MULTIEXPR_SUBLINK   UPDATE-only path, also blocked by the
+				 *                       PARAM_MULTIEXPR guard above.
+				 *   ARRAY_SUBLINK       ARRAY() construction differs in CH.
+				 *   CTE_SUBLINK         Recursive/materialised CTE survives
+				 *                       past inline_cte_walker; CH cannot host.
```
theory, line 428: What does this mean? Possible to support in the future? What would it take?

src/deparse.c
```diff
@@ -380,6 +403,101 @@ foreign_expr_walker(Node * node,
...
+				 * Supported SubLinkTypes:
+				 *   ANY_SUBLINK    -> col IN (SELECT ...)   (equality only)
+				 *   EXISTS_SUBLINK -> EXISTS (SELECT ...)
+				 *   EXPR_SUBLINK   -> (SELECT ...)
+				 *
+				 * Unsupported (and why):
+				 *   ALL_SUBLINK         CH lacks ALL; would need NOT EXISTS
+				 *                       rewrite. NOT IN already arrives as
+				 *                       NOT(ANY_SUBLINK).
+				 *   ROWCOMPARE_SUBLINK  Multi-column compare not deparsed.
```
theory, line 423: Could they be? What would need to change?

serprex: they could be, but will require more intrusive deparsing logic, on CH I think it could work by converting to using tuples. but seems best to have initial PR manage dealing with pg subplan logic where syntax is pretty 1:1, then followup can expand

src/deparse.c
serprex: we're operating on AST after some optimizations, CTE_SUBLINK at this point is complicated enough to not be optimized

CH does not support recursive CTE: https://clickhouse.com/docs/sql-reference/statements/select/with

> Recursion is prevented by hiding the current CTE from the identifier resolution process.

theory: Yeah, would be nice to support non-recursive CTE in the future.

serprex:
failed experiment, may be used as a reference by some future attempt

PR #188, opened 2026-04-05
Fix parser to not expect well formed input

serprex:
split out from #171

### DIFF
--- a/src/parser.c
+++ b/src/parser.c
@@ -125,8 +125,11 @@ ch_http_read_next(ch_http_read_state * state)
 		return CH_CONT;
 	}
 
-	/* Should be at the end of the line. */
-	Assert(data[state->curpos] == '\n');
+	/* Should be at the end of the line or the file. */
+	if (data[state->curpos] != '\n')
+	{
+		elog(ERROR, "Unexpected byte (%d) after array", data[state->curpos]);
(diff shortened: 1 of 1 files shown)


src/parser.c
```diff
@@ -92,8 +92,11 @@ ch_http_read_next(ch_http_read_state * state)
 	}
 
 	/* Should be at the end of the line or the file. */
-	Assert(data[state->curpos] == '\n');
-	int			res = data[state->curpos + 1] == '\0' ? CH_EOF : CH_EOL;
+	if (data[state->curpos] != '\n' && data[state->curpos] != '\0')
+	{
+		elog(ERROR, "Unexpected character %c after array", data[state->curpos]);
+	}
+	int			res = data[state->curpos] == '\0' || data[state->curpos + 1] == '\0' ? CH_EOF : CH_EOL;
```
theory: Is there a test we can add for this?

theory (approved):

src/parser.c
serprex: bit hard since it'd require a misbehaving clickhouse instance. if we had something like mitmproxy in CI it could force it

theory: Fair enough; created #194 to capture the general idea.

PR #190, opened 2026-04-05
Support more array functions

serprex:
Builtin array functions were presumed shippable, emitted with PG names, failing on ClickHouse.
Map to CH equivalents, mark unmappable as unshippable
Previous subscript code passed `arr[x:y]` to ClickHouse as-is, which is unsupported syntax

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,14 @@ All notable changes to this project will be documented in this file. It uses the
     prepending them to the regular expression (e.g., `(?i)foo`).
 *   Added pushdown for `regexp_split_to_array()` to `splitByRegexp()`,
     including flags.
+*   Added pushdown mappings for array functions: `array_cat`, `array_append`,
+    `array_remove`, `array_to_string`, `cardinality`, `array_length`,
+    `array_prepend`, `string_to_array`, `trim_array`, `array_fill`,
+    `array_reverse`, `array_shuffle`, `array_sample`, `array_sort`.
+*   Added pushdown for array operators: `@>` (`hasAll`), `<@` (`hasAll`),
+    `&&` (`hasAny`).
+*   Array slice syntax (`arr[L:U]`, `arr[:U]`, `arr[L:]`) now pushes down
+    as `arraySlice()`.
 
 ### ⬆️ Dependency Updates
 
@@ -39,6 +47,10 @@ All notable changes to this project will be documented in this file. It uses the
     unable to map a ClickHouse type to a Postgres type.
 *   Fixed reversal of the arguments passed to the ClickHouse `match()`
     function by the mapping from `regexp_like()`.
+*   `array_dims`, `array_ndims`, `array_lower`, `array_upper`, `array_replace`,
+    `array_positions`, `array_fill (3-arg)`, `array_sort (3-arg)`, and
+    `string_to_array(3-arg)` now evaluate locally instead of being pushed to
+    ClickHouse where they would fail.
 
   [v0.1.11]: https://github.com/ClickHouse/pg_clickhouse/compare/v0.1.10...v0.1.11
 
--- a/doc/pg_clickhouse.md
+++ b/doc/pg_clickhouse.md
@@ -287,13 +287,13 @@ Use [CREATE FOREIGN TABLE] to create a foreign table that can query data from
 a ClickHouse database:
 
 ```sql
-CREATE FOREIGN TABLE uact (
+CREATE FOREIGN TABLE acts (
     user_id    bigint NOT NULL,
     page_views int,
     duration   smallint,
     sign       smallint
 ) SERVER taxi_srv OPTIONS(
-    table_name 'uact',
+    table_name 'acts',
     engine 'CollapsingMergeTree'
 );
 ```
@@ -325,8 +325,6 @@ option:
 
 Example:
 
-(aggregatefunction 'sum')
-
 ```sql
 CREATE FOREIGN TABLE test (
     column1 bigint  OPTIONS(AggregateFunction 'uniq'),
@@ -354,14 +352,14 @@ TABLE].
 Use [DROP FOREIGN TABLE] to remove a foreign table:
 
 ```sql
-DROP FOREIGN TABLE uact;
+DROP FOREIGN TABLE acts;
 ```
 
 This command fails if there are any objects that depend on the foreign table.
 Use the `CASCADE` clause to drop them, too:
 
 ```sql
-DROP FOREIGN TABLE uact CASCADE;
+DROP FOREIGN TABLE acts CASCADE;
 ```
 
 ## DML SQL Reference
@@ -422,7 +420,7 @@ like any other tables:
 try=# SELECT start_at, duration, resource FROM logs WHERE req_id = 4117909262;
           start_at          | duration |    resource
 ----------------------------+----------+----------------
- 2025-12-05 15:07:32.944188 |      175 | /widgets/totam
+ 2025-12-05 15:07:32.944188 |      175 | /widgets/totem
 (1 row)
 ```
 
@@ -922,7 +920,7 @@ SELECT c1, encode(c2::bytea, 'hex'), encode(c3::bytea, 'hex') FROM texts ORDER B
 
 The text columns will be correct:
 
-```pgdsql
+```pgsql
 
  c1 |                          encode                          |              encode
 ----+----------------------------------------------------------+----------------------------------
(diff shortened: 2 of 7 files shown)


doc/pg_clickhouse.md
```diff
@@ -1021,6 +1021,23 @@ maps the following functions:
...
+*   `array_fill`: [arrayWithConstant](https://clickhouse.com/docs/sql-reference/functions/array-functions#arrayWithConstant)
+*   `array_reverse`: [arrayReverse](https://clickhouse.com/docs/sql-reference/functions/array-functions#arrayReverse)
+*   `array_shuffle`: [arrayShuffle](https://clickhouse.com/docs/sql-reference/functions/array-functions#arrayShuffle)
+*   `array_sample`: [arrayRandomSample](https://clickhouse.com/docs/sql-reference/functions/array-functions#arrayRandomSample)
+*   `array_sort`: [arraySort](https://clickhouse.com/docs/sql-reference/functions/array-functions#arraySort) / [arrayReverseSort](https://clickhouse.com/docs/sql-reference/functions/array-functions#arrayReverseSort)
+*   Array slice (`arr[L:U]`): [arraySlice](https://clickhouse.com/docs/sql-reference/functions/array-functions#arraySlice)
+*   `@>` (array contains): [hasAll](https://clickhouse.com/docs/sql-reference/functions/array-functions#hasAll)
+*   `<@` (array contained by): [hasAll](https://clickhouse.com/docs/sql-reference/functions/array-functions#hasAll)
+*   `&&` (array overlap): [hasAny](https://clickhouse.com/docs/sql-reference/functions/array-functions#hasAny)
```
theory: I'd like to add a separate section for operators.

src/custom_types.c
```diff
@@ -209,6 +229,42 @@ chfdw_check_for_custom_function(Oid funcid)
...
+			case F_ARRAY_DIMS:
+			case F_ARRAY_NDIMS:
+			case F_ARRAY_LOWER:
+			case F_ARRAY_UPPER:
+			case F_ARRAY_REPLACE:
+			case F_ARRAY_POSITIONS:
+			case F_ARRAY_TO_STRING_ANYARRAY_TEXT_TEXT:
+			case F_STRING_TO_ARRAY_TEXT_TEXT_TEXT:
+			case F_ARRAY_FILL_ANYELEMENT__INT4__INT4:
+#if PG_VERSION_NUM >= 180000
```
theory: Are these `#if PG_VERSION_NUM >= 180000`s necessary? Seems like we can just add the missing constants above as usual, no?

src/custom_types.c
```diff
@@ -414,6 +541,9 @@ static const BuiltinOperatorMap builtin_operator_map[] = {
 	{"!~*", CF_REGEX_ICASE_NO_MATCH},
 	{"->", CF_JSONB_FETCHVAL},
 	{"->>", CF_JSONB_FETCHVAL_TEXT},
+	{"@>", CF_ARRAY_CONTAINS},
+	{"<@", CF_ARRAY_CONTAINED_BY},
+	{"&&", CF_ARRAY_OVERLAP},
```
theory: We're going to run into problems matching operators with strings rather than constants, because they can be used for different data types. For example, `@>` can also be used for JSON intersection as well as array overlap. I think we should fix this or there will be weird errors.

src/deparse.c
```diff
@@ -2612,14 +2623,56 @@ deparseFuncExpr(FuncExpr * node, deparse_expr_cxt * context)
...
-		deparseExpr((Expr *) list_nth(node->args, 1), context);
-		appendStringInfoString(buf, ", ");
-		deparseExpr((Expr *) linitial(node->args), context);
-		appendStringInfoChar(buf, ')');
-		return;
-	}
+	if (cdef)
+		switch (cdef->cf_type)
+		{
+			case CF_TIMEZONE:
```
theory: This will conflict with 970edeb, which makes the same change.

test/expected/array_functions.out
```diff
@@ -0,0 +1,474 @@
...
+$$);
+ clickhouse_raw_query 
+----------------------
+ 
+(1 row)
+
+CREATE SCHEMA arr_test;
+IMPORT FOREIGN SCHEMA arr_test FROM SERVER arr_svr INTO arr_test;
+SET search_path = arr_test, public;
+-- Simple mappings
```
theory: Pushdown not exercised in these queries.

test/expected/array_functions.out
```diff
@@ -0,0 +1,474 @@
...
+   Remote SQL: SELECT NULL FROM arr_test.t1
+(3 rows)
+
+EXPLAIN (VERBOSE, COSTS OFF)
+SELECT * FROM t1 WHERE trim_array(vals, 1) = ARRAY[10,20];
+                                                   QUERY PLAN                                                   
+----------------------------------------------------------------------------------------------------------------
+ Foreign Scan on arr_test.t1
+   Output: id, vals, tags
+   Remote SQL: SELECT id, vals, tags FROM arr_test.t1 WHERE (((arrayResize(vals, length(vals) - 1)) = [10,20]))
```
theory: First query to exercise pushdown. Can we see a result set, too? Same for the other pushdown queries below. We need to see that ClickHouse actually processes then correctly and there are no syntax errors.

test/expected/array_functions.out
```diff
@@ -0,0 +1,474 @@
...
+ {40}
+ {}
+(3 rows)
+
+SELECT array_fill(42, ARRAY[3]) FROM t1 LIMIT 1;
+ array_fill 
+------------
+ {42,42,42}
+(1 row)
+
```
theory: These tests just exercise the Postgres functions; there is no need for them IMO. Same for the others that don't pushdown.

theory (changes requested):

src/custom_types.c
serprex: converted all operator logic to use oids too

src/custom_types.c
serprex: yes, waffling on best way here

test/expected/array_functions.out
serprex: in select expressions these don't get passed down, moved to WHERE

src/custom_types.c
theory: For previous cases I kept the conditionals to settling the constants and didn't worry about them here.

doc/pg_clickhouse.md
```diff
@@ -1055,6 +1068,13 @@ maps the following functions:
 *   `CURRENT_ROLE`: Passed as value from PostgreSQL function.
 *   `SESSION_USER`: Passed as value from PostgreSQL function.
 
+### Pushdown Operators
+
+*   Array slice (`arr[L:U]`): [arraySlice](https://clickhouse.com/docs/sql-reference/functions/array-functions#arraySlice)
+*   `@>` (array contains): [hasAll](https://clickhouse.com/docs/sql-reference/functions/array-functions#hasAll)
+*   `<@` (array contained by): [hasAll](https://clickhouse.com/docs/sql-reference/functions/array-functions#hasAll)
+*   `&&` (array overlap): [hasAny](https://clickhouse.com/docs/sql-reference/functions/array-functions#hasAny)
+
```
theory, line 1082: There are others; Kaushik added regex operators last week. We should document them, as well.

src/custom_types.c
```diff
@@ -349,6 +398,69 @@ chfdw_check_for_custom_function(Oid funcid)
 					entry->custom_name[0] = '\1';
 					break;
 				}
+				/* array functions: unshippable */
```
theory, line 415: Presumably over time we'll add more unshippables. Perhaps they should be listed last…

src/deparse.c
```diff
@@ -2508,8 +2519,20 @@ deparseFuncExpr(FuncExpr * node, deparse_expr_cxt * context)
...
+	if (cdef && cdef->cf_type == CF_ARRAY_SORT_DESC)
+	{
+		Expr	   *desc_arg = (Expr *) list_nth(node->args, 1);
+
+		if (IsA(desc_arg, Const) && !((Const *) desc_arg)->constisnull
+			&& DatumGetBool(((Const *) desc_arg)->constvalue))
+			strcpy(cdef->custom_name, "arrayReverseSort");
+		else
+			strcpy(cdef->custom_name, "arraySort");
+	}
```
theory, line 2564: Do we not have enough information to do this in `chfdw_check_for_custom_function()`?

src/deparse.c
```diff
@@ -2652,14 +2675,46 @@ deparseFuncExpr(FuncExpr * node, deparse_expr_cxt * context)
...
-	if (cdef && (cdef->cf_type == CF_TIMEZONE || cdef->cf_type == CF_MATCH))
-	{
-		deparseExpr((Expr *) list_nth(node->args, 1), context);
-		appendStringInfoString(buf, ", ");
-		deparseExpr((Expr *) linitial(node->args), context);
-		appendStringInfoChar(buf, ')');
-		return;
-	}
+	if (cdef)
+		switch (cdef->cf_type)
```
theory: Holy crap, I didn't realize a switch was considered a single expression so the `if` doesn't require parens. 🤯

test/expected/array_functions.out
```diff
@@ -0,0 +1,391 @@
...
+-- array_remove → arrayRemove (CH 26+, EXPLAIN only)
+EXPLAIN (VERBOSE, COSTS OFF)
+SELECT * FROM t1 WHERE array_remove(vals, 20) = ARRAY[10,30];
+                                           QUERY PLAN                                           
+------------------------------------------------------------------------------------------------
+ Foreign Scan on arr_test.t1
+   Output: id, vals, tags
+   Remote SQL: SELECT id, vals, tags FROM arr_test.t1 WHERE ((arrayRemove(vals, 20) = [10,30]))
+(3 rows)
```
theory, line 82: Missing actual execution for this one.

test/expected/array_functions.out
```diff
@@ -0,0 +1,391 @@
...
+   Remote SQL: SELECT id, vals, tags FROM arr_test.t1 WHERE ((arraySlice(vals, 1) = [10,20,30]))
+(3 rows)
+
+SELECT * FROM t1 WHERE vals[:] = ARRAY[10,20,30];
+ id |    vals    |  tags   
+----+------------+---------
+  1 | {10,20,30} | {a,b,c}
+(1 row)
+
+-- Unshippable (function NOT in Remote SQL)
```
theory, line 334: There are a bunch more of these.

CHANGELOG.md
```diff
@@ -24,6 +24,14 @@ All notable changes to this project will be documented in this file. It uses the
     This is the same value visible in `pg_get_loaded_modules()`, but available
     in Postgres versions prior to 18, and without having to load
     pg_clickhouse, first.
+*   Added pushdown mappings for array functions: `array_cat`, `array_append`,
+    `array_remove`, `array_to_string`, `cardinality`, `array_length`,
+    `array_prepend`, `string_to_array`, `trim_array`, `array_fill`,
+    `array_reverse`, `array_shuffle`, `array_sample`, `array_sort`.
+*   Added pushdown for array operators: `@>` (`hasAll`), `<@` (`hasAll`),
+    `&&` (`hasAny`).
+*   Array slice syntax (`arr[L:U]`, `arr[:U]`, `arr[L:]`) now pushes down
```
theory, line 37: The subscript syntax is new, too, no?

theory (approved):
This is great. Just a few tweaks requested. Also please provide a detailed explanation of the changes in the commit message. Goals is to keep the Git history as the canonical history, rather than GitHub PRs. TIA

CHANGELOG.md
serprex: new how? previously it was being pushed down as-is, which is invalid CH syntax

theory: [Is this not new behavior](https://github.com/ClickHouse/pg_clickhouse/pull/190/changes#diff-0444004ab17cf299ec448ddfd6a53ab644da711622e4560e44a1d4b1bd30df20R278-R292)?

serprex: yes, hence in changelog, confused

theory: The changelog mentions slice syntax, not single-item subscript. Perhaps that's the same as slice? I'd expect something like

```
*   Array subscript syntax (`arr[n]`) and slice syntax (`arr[L:U]`, `arr[:U]`, `arr[L:]`) now pushes down
```

serprex: array subscript was supported, as that syntax is compatible

src/deparse.c
serprex: should also make it unshippable when sort order is non-constant

theory: I'll like to make this change in a subsequent PR.

PR #202, opened 2026-04-08
Add mapping for `regexp_split_to_array()` + flags

theory:
Map the Postgres `regexp_split_to_array()` function to ClickHouse `splitByRegexp()`. When flags are passed as the optional third argument to `regexp_split_to_array()`, prepend them in `(?)` to the regular expression.

While at it, do the same for `regexp_like()` flags, and fix the bug that reversed the arguments to the ClickHouse `match()` function, which has existed since the mapping was introduced.

Refactor `deparseFuncExpr()` to handle the the `toTimeZone()`, `match()`, and `splitByRegexp()` cases separately, but abstract creating the regular expression with optional flags into the new `appendRegex()` function. Alas, ClickHouse reverses the arguments to `match()` and `splitByRegexp()`, so we can't treat them the same.

Add tests for these behaviors to `functions.sql`, and add some notes on differences in regular expression behavior and flags to the reference documentation.

### DIFF
--- a/doc/pg_clickhouse.md
+++ b/doc/pg_clickhouse.md
@@ -1032,6 +1032,7 @@ maps the following functions:
 *   `btrim`: [trimBoth](https://clickhouse.com/docs/sql-reference/functions/string-functions#trimboth)
 *   `strpos`: [position](https://clickhouse.com/docs/sql-reference/functions/string-search-functions#position)
 *   `regexp_like`: [match](https://clickhouse.com/docs/sql-reference/functions/string-search-functions#match)
+*   `regexp_split_to_array`: [splitByRegexp](https://clickhouse.com/docs/sql-reference/functions/splitting-merging-functions#splitByRegexp)
 *   `md5`: [MD5](https://clickhouse.com/docs/sql-reference/functions/hash-functions#MD5)
 *   `to_timestamp(float8)`: [fromUnixTimestamp](https://clickhouse.com/docs/sql-reference/functions/date-time-functions#fromUnixTimestamp)
 *   `statement_timestamp`, `transaction_timestamp`, & `clock_timestamp`:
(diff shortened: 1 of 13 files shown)


doc/pg_clickhouse.md
```diff
@@ -1159,6 +1160,16 @@ rejects frame specifications on these functions.
...
 
+## Compatibility Notes
+
+### Regular Expressions
+
+While pg_clickhouse pushes down regular expressions to ClickHouse equivalents,
+be aware of the differences between the two.
+
+*   PostgreSQL supports [POSIX Regular Expressions]
+*   PostgreSQL [Regex flags]
```
theory: In progress, will flesh out once I find the ClickHouse regex docs.

serprex:
I'm of the opinion we shouldn't be pushing this stuff down

ClickHouse regex is RE2 https://clickhouse.com/docs/interfaces/formats/Regexp#format-settings
whereas Postgres is some POSIX with a bit of Perl, https://www.postgresql.org/docs/current/functions-matching.html#POSIX-SYNTAX-DETAILS

https://github.com/google/re2/wiki/syntax
RE2 doesn't support backreferences

https://github.com/ClickHouse/ClickHouse/blob/c1225a670d2e98c072f97d20347b7d57de805602/src/Functions/match.cpp#L31
CH has `.` match newline by default, `s` option disables, whereas postgres has `.` not match newline by default, `s` option enables

word boundaries use different classes (b/B on RE2, m/M/y/Y on postgres)

...the list goes on

Seems like correct pushdown needs to validate a shared subset of patterns between their dialects, possibly with basic transforms. As is a user must know exactly whether their regex is being pushed down or not & write their query accordingly. This is where I'd want to have RE2 regex UDFs exposed to postgres based on ClickHouse's rules which users can have function on postgres & pushdown 1:1 on pg_clickhouse

iskakaushik (approved):

serprex:
to summarize decisions:

* merge this PR as is
* introduce GUC to enable/disable regex pushdown (on by default)
* we will create `pg_re2ch` for re2regex-in-postgres extension that can be 1:1 pushed down
* in future if we can validate safe regex pushdown for basic regex, we can change GUC to default off

PR #204, opened 2026-04-10
Support pushdown for re2 extension

serprex:
ClickHouse is working on a postgres extension using re2 for regex

This addresses incompatibility between CH & PG regex,
where engine depends on whether function is pushed down or not

### DIFF
--- a/doc/pg_clickhouse.md
+++ b/doc/pg_clickhouse.md
@@ -1105,6 +1105,34 @@ any of these functions cannot be pushed down they will raise an exception.
 
 *   [dictGet](https://clickhouse.com/docs/sql-reference/functions/ext-dict-functions#dictget-dictgetordefault-dictgetornull)
 
+### Extension Pushdown
+
+pg_clickhouse recognizes functions from select core and third-party extensions,
+pushing them down to their ClickHouse equivalents.
+
+#### re2
+
+All [re2 extension] functions push down 1:1 to ClickHouse:
+
+*   `re2match` → [match](https://clickhouse.com/docs/sql-reference/functions/string-search-functions#match)
+*   `re2extract` → [extract](https://clickhouse.com/docs/sql-reference/functions/string-search-functions#extract)
+*   `re2extractall` → [extractAll](https://clickhouse.com/docs/sql-reference/functions/string-search-functions#extractAll)
+*   `re2regexpextract` → [regexpExtract](https://clickhouse.com/docs/sql-reference/functions/string-search-functions#regexpExtract)
+*   `re2extractgroups` → [extractGroups](https://clickhouse.com/docs/sql-reference/functions/string-search-functions#extractGroups)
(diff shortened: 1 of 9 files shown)


test/expected/result_map.txt
```diff
@@ -225,3 +225,16 @@ window_functions.sql
...
+
+ Postgres | pg_re2    | File
+----------|-----------|----------------------
+ 16-18    | installed | re2_functions.out
+ 18       | absent    | re2_functions_1.out
+ 17       | absent    | re2_functions_6.out
+ 16       | absent    | re2_functions_5.out
+ 15       | absent    | re2_functions_2.out
+ 14       | absent    | re2_functions_4.out
+ 13       | absent    | re2_functions_3.out
```
theory: Might be better to detect whether the extension is installed in the test and skip them all if it's not. Then need only two output files.

doc/pg_clickhouse.md
```diff
@@ -1323,3 +1349,7 @@ Copyright (c) 2025-2026, ClickHouse.
   [Postgres flags]: https://www.postgresql.org/docs/18/functions-matching.html#POSIX-EMBEDDED-OPTIONS-TABLE
     "PostgreSQL Docs: ARE Embedded-Option Letters"
   [RE2 Regular Expressions]: https://github.com/google/re2/wiki/Syntax "RE2 Syntax"
+  [pg_re2]: https://github.com/ClickHouse/pg_re2
+    "pg_re2: ClickHouse-compatible regex functions using RE2"
```
theory, line 1392: Are they still ClickHouse-specific?

doc/pg_clickhouse.md
```diff
@@ -1088,6 +1088,32 @@ any of these functions cannot be pushed down they will raise an exception.
 
 *   [dictGet](https://clickhouse.com/docs/sql-reference/functions/ext-dict-functions#dictget-dictgetordefault-dictgetornull)
 
+### Extension Pushdown
+
+pg_clickhouse recognizes functions from select third-party extensions and
+pushes them down to their ClickHouse equivalents.
```
```suggestion
pg_clickhouse recognizes functions from select core and third-party extensions
and pushes them down to their ClickHouse equivalents.
```

theory: Document support for intarray `idx()` function.

theory (commented):
Let's add `re2` as a [recommended prereq](https://pgxn.org/spec/#prereqs):

```patch
--- a/META.json
+++ b/META.json
@@ -16,6 +16,9 @@
     "runtime": {
       "requires": {
         "PostgreSQL": "13.0.0"
+      },
+      "recommends": {
+        "pgre2": "0.1.0"
       }
     }
   },
```

A few other suggestions; would be nice to simplify the expected test output.

doc/pg_clickhouse.md
serprex: we set `opts.set_dot_nl(true)` to be ClickHouse compatible. & functions match ClickHouse function signatures by design

theory (approved):

PR #207, opened 2026-04-10
Improve `regexp_replace()` pushdown

theory:
Previously pg_clickhouse pushed down `regexp_replace()` without modification, which would work for simple cases. However, `regexp_replace()` in ClickHouse is an alias for `replaceRegexpAll()`, but Postgres `regexp_replace()` doesn't replace all unless the `g` flag is passed.

Add an explicit mapping for `regexp_replace(text, text, text)` and `regexp_replace(text, text, text, text)`. In `deparseFuncExpr`, determine whether flags are present and, if so, parse and remove any and all `g`s. If there are any, push down to `replaceRegexpAll()`; otherwise push down to `replaceRegexOne()` to match the default Postgres behavior.

As for the other supported regular expression functions, prepend the flags (sans `g`s) to the matching string as `(?$flags)`.

Add tests for these behaviors.

### DIFF
--- a/doc/pg_clickhouse.md
+++ b/doc/pg_clickhouse.md
@@ -1044,6 +1044,7 @@ maps the following functions:
 *   `btrim`: [trimBoth](https://clickhouse.com/docs/sql-reference/functions/string-functions#trimboth)
 *   `strpos`: [position](https://clickhouse.com/docs/sql-reference/functions/string-search-functions#position)
 *   `regexp_like`: [match](https://clickhouse.com/docs/sql-reference/functions/string-search-functions#match)
+*   `regexp_replace`: [replaceRegexpOne](https://clickhouse.com/docs/sql-reference/functions/string-replace-functions#replaceRegexpOne) or [replaceRegexpOne](https://clickhouse.com/docs/sql-reference/functions/string-replace-functions#replaceRegexpAll) when the `g` flag is present
 *   `regexp_split_to_array`: [splitByRegexp](https://clickhouse.com/docs/sql-reference/functions/splitting-merging-functions#splitByRegexp)
 *   `md5`: [MD5](https://clickhouse.com/docs/sql-reference/functions/hash-functions#MD5)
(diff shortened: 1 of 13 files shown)


doc/pg_clickhouse.md
```diff
@@ -1210,17 +1212,26 @@ be aware of the differences between the two.
...
-  The only flags both support, and therefore can be used when evaluated by
-  ClickHouse, are:
+*   The only flags both support, and therefore can be used when evaluated by
+    ClickHouse, are:
+
+    RE2 supports only these flags; don't use any other [Postgres flags]
+
+    *   `i`: case-insensitive
+    *   `m`: multi-line mode:
+    *   `s`: let `.` match `\n`
```
serprex, line 1222: `s` is weird. In ClickHouse `.` matches `\n` by default, you need to pass `s` to change this. In Postgres you'd pass `p` or `n` for that behavior *(I have not tested ^/$ matching rules)*

https://www.postgresql.org/docs/current/functions-matching.html (ctrl-f for "newline-sensitive matching")

https://clickhouse.com/docs/sql-reference/functions/string-search-functions#match

> Unlike re2's default behavior, . matches line breaks. To disable this, prepend the pattern with (?-s)

theory: Oh WTF flags support `-`? Sheesh.

theory: Since the Postgres parser will reject `-`, perhaps we should pass `-s` if there is no `s`.

serprex: yes, & don't pass `-s` if `p` flag specified (ie, ignore `p`)

theory: Okay, working on:

* Allow `s`, `i`, `m`
* Append `-s` unless `s` or `p`
* Remove `t` and `p`
* Disallow anything else

Waiting to hear from ClickHouse devs whether this is only relevant to `match()`.

theory: Working on this for the next commit, BTW, would like to keep this one as-is.

serprex (approved):

PR #208, opened 2026-04-10
Validate and transform regular expression flags

theory:
The ClickHouse [match docs] say:

> Unlike re2's default behavior, `.` matches line breaks. To disable this, prepend the pattern with `(?-s)`.

To match the default behavior of Postgres regular expression operators and functions, always inject `(?-s)`. This requires examining the flags argument to the regular expression functions and only injecting it if no `s` is present.

The new function `regex_flags_ok()` examines these flags, returning false for any flags not supported by ClickHouse. `chfdw_is_shippable` thus marks such function calls as unshippable if they use flags that cannot be pushed down. It also returns false if the flags come from a dynamic argument, since such an argument cannot be pushed down, either. An additional note suggests also parsing the regular expression text for additional flags to carry out the same process.

Update `appendRegex` to call yet another new function, `appendRegexFlags()`, which parses the flags passed to a regex function and appends `-s` if no `s` is present.

Also update the handling of regular expression operators to always insert `-s`.

Move the regular expression function tests to their own test file, `regex`, in order to avoid all the complexity of the `functions` tests. Use a `DO` block to test `regexp_like()` only on Postgres 15+.

Unfortunately, ClickHouse 24 returns unexpected results when using `(?-s)` in a few test cases, so we still need a `regex_1.out` to handle that case, as well as the new `ilike_regex.out`, for the same reason.

While about things, rename the constants ending in `_REGEXP` to ``_REGEX` to match other regular expression constants.

  [match docs]: https://clickhouse.com/docs/sql-reference/functions/string-search-functions#match


test/expected/ilike_regex_1.out
```diff
@@ -0,0 +1,433 @@
...
+   Output: (count(*))
+   Relations: Aggregate on (events)
+   Remote SQL: SELECT count(*) FROM ilike_regex_test.events WHERE (match(name, concat('(?-s)', '_view$')))
+(4 rows)
+
+SELECT count(*) FROM ilike_regex_bin.events WHERE name ~ '_view$';
+ count 
+-------
+     0
+(1 row)
```
theory, line 302: This is the unexpected change on ClickHouse 24. It returns 1 for `match(name, '_view$')`. Weirdly, `match(name, '(?s)_view$')` also fails.

test/expected/regex_1.out
```diff
@@ -0,0 +1,180 @@
...
+---------------------------------------------------------------------------------------------------------------------
+ Foreign Scan on public.strings
+   Output: val
+   Remote SQL: SELECT val FROM regex_test.strings WHERE ((splitByRegexp(concat('(?-s)', ','), val) = ['a','b','c']))
+(3 rows)
+
+SELECT val FROM strings WHERE regexp_split_to_array(val, ',') = '{a,b,c}'::text[];
+ val 
+-----
+(0 rows)
```
theory, line 46: This is the one test that returns a different result on ClickHouse 24.

serprex (approved):

PR #209, opened 2026-04-11
Convert much elog to ereport

serprex:


theory (commented):
I kinda think they should all have the `pg_clickhouse: ` prefix, too, but this is an improvement as-is.

theory (approved):

PR #213, opened 2026-04-13
Remove Debian workflow

theory:
Will integrate it into the release workflow at a later date.


iskakaushik (approved):

ibotty:
I was looking forward to consume the packages.  Would you consider a PR that adds the debian package building to the release workflow?

theory:
@ibotty Of course! Might have to borrow some stuff from this pattern, since the pgxn-tools container is currently amd64-only. My ideal pattern is that the logic lives in a shell script and can be run by a `make` command, and that's all the workflow has to do. Makes it easier to test and maintain it.

ibotty:
I also prefer having having a Makefile that builds the package.  Should it still build the extension separately or copy the built extension (from other stages of the release workflow)?  Or should all/most of the release workflow be part of the Makefile?

Also:  do you know how to submit the package to apt.postgresql.org?  Would you be willing to publish the extension there?

theory:
> Should it still build the extension separately or copy the built extension (from other stages of the release workflow)?

I don't think it matters, really. Nice if you can copy it, but as I said, the pgxn-tools-based pipelines only build for amd64 at the moment.

> Or should all/most of the release workflow be part of the Makefile?

As much as possible IMHO.

> Also: do you know how to submit the package to apt.postgresql.org? Would you be willing to publish the extension there?

Yes, but it's a separate process. @df7cb would have to build it (or take a PR for the [apt.postgresql.org repo](https://salsa.debian.org/postgresql/apt.postgresql.org). There a currently a few wrinkles, mainly the vendered dependencies.

df7cb:
Yeah looking at pg_clickhouse is still on my todo list, but work stuff and vacations had priority until now.

PR #220, opened 2026-04-18
Add failing streaming test

theory:
Set the fetch size to just one byte, then have it fetch 3500 integers. At a certain point, it fails with

    ERROR:  invalid input syntax for type integer: "Host: 127.0.0.1:8123

Seems like somehow a header is fed into the fetch routine. Seen on macOS with libcurl 8.19.0.


theory:
Resolved in #267.

PR #222, opened 2026-04-18
Unify HTTP transfer engine into the streaming driver

iskakaushik:
## Summary
- `ch_http_simple_query` is now a thin wrapper over `ch_http_stream_begin` with `fetch_size=INT32_MAX` — no more duplicated URL/settings/MIME setup across the two HTTP paths
- Added `ch_http_stream_take_body` to transfer ownership of the response buffer to the caller (zero-copy)

## Test plan
- [x] `make tempcheck` — all 27 pg_regress tests pass (including `http`, `http_inserts`, `query_cancel`, `param`, and binary/FDW cross-path tests that exercise every caller)
- [x] Clean build, zero warnings

### DIFF
--- a/src/http_streaming.c
+++ b/src/http_streaming.c
@@ -530,3 +530,44 @@ ch_http_stream_total_time(HttpStream * stream)
 {
 	return stream->total_time * 1000;
 }
+
+/*
+ * ch_http_stream_take_body — transfer ownership of the response body.
+ *
+ * On return, *out_data is a malloc()'d buffer the caller must free(). When
+ * status is CH_HTTP_STATUS_TRANSPORT_ERROR the body is the strdup'd libcurl
+ * error message; otherwise it is the accumulated response bytes (shifted to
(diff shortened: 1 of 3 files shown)


src/http_streaming.c
```diff
@@ -530,3 +530,33 @@ ch_http_stream_total_time(HttpStream * stream)
 {
 	return stream->total_time * 1000;
 }
+
+void
+ch_http_stream_take_body(HttpStream * stream, char **out_data, size_t *out_size)
```
theory: Would you add doc comments, please?

theory (approved):
Very happy to be eliminating the redundancy. Wondering if, in a future PR, we should consider using Postgres memory contexts rather than `malloc()`.

src/http_streaming.c
iskakaushik: Added a doc comment spelling out ownership, NUL-termination, and one-shot semantics in 65a5edc. The header still has the short version; this one lives with the implementation.

PR #223, opened 2026-04-19
Fix EXPLAIN (VERBOSE) for window functions

serprex:


src/fdw.c
```diff
@@ -771,6 +845,26 @@ clickhouseGetForeignPlan(PlannerInfo * root,
 	/* Remember remote_exprs for possible use by clickhousePlanDirectModify */
 	fpinfo->final_remote_exprs = remote_exprs;
 
+	/*
+	 * When window functions are pushed down, ForeignScan takes the place of
+	 * what would otherwise be a WindowAgg plan node. Core EXPLAIN VERBOSE
+	 * deparses WindowFunc via a WindowClause (query context) or WindowAgg in
+	 * the plan-namespace stack; neither exists here, so deparse errors with
+	 * "could not find window clause for winref N". Remote SQL has already
```
theory: Is it an error that's departed? Seems like we avoid it.

test/sql/window_functions.sql
```diff
@@ -237,6 +237,71 @@ WHERE event_name = 'lead_created'
 ORDER BY entity_id, ts_event
 LIMIT 3;
 
+-- ============================================================
+-- EXPLAIN VERBOSE with pushed-down window functions
+-- Regression: deparse of WindowFunc in fdw_scan_tlist used to fail
+-- with "could not find window clause for winref N".
+-- ============================================================
```
theory, line 244: Why not just update all the above `EXPLAIN`s to be verbose? This seems redundant.

theory (approved):
This is great, thank you. Am I wrong in thinking that it's kind of a hack, though? I wonder if the core needs more tooling around this sort of thing.

test/sql/window_functions.sql
serprex: at least for PR there's value in seeing that this change has no impact on anything else. also good to cover various explain flavors

src/fdw.c
serprex: right, will update with tense "would error"

serprex:
yes, this is absolutely a hack, unfortunately postgres deparsing lends itself to being hacked around for FDWs

PR #226, opened 2026-04-25
test: add misbehaving mock ClickHouse harness (#194)

MukundaKatta:
## Summary
Closes #194. Adds an in-process Python `ThreadingHTTPServer` mock that returns malformed, partial, slow, or dropped HTTP responses, plus a 17-case pytest suite that locks down the wire-level contract of each misbehavior.

## Mock approach (`test/mock/mock_clickhouse.py`)
Dispatches on a `-- BEHAVIOR=<name>` marker embedded in the request body or `?behavior=` query string. Each behavior emits malformed bytes directly to the socket — some bypass `send_response` / `shutdown` mid-body — so the real `libcurl` + `parser.c` paths see the kinds of responses real ClickHouse never produces.

## Behaviors covered
5xx errors with and without `X-ClickHouse-Exception-Code`, unterminated array literals (PR #188 regression class), truncated chunked transfer, slow headers / timeout exercise, truncated JSON, mid-body drops, trailing binary garbage, unterminated quoted strings, backslash at EOF, wrong `Content-Length`, request log.

## Files changed
All new, all under `test/mock/`:
- `mock_clickhouse.py` — server
- `test_mock_clickhouse.py` — pytest (17 cases)
- `conftest.py` — sys.path
- `mock_misbehaving.sql` — opt-in SQL driver, deliberately outside `test/sql/` so `make installcheck` is unaffected
- `README.md` — documentation

761 lines total. **No source code changes.**

## Test plan
- `ruff format` clean, `ruff check` clean, `codespell` clean
- `pytest -v` 17/17 passed in 8.26s
- Verifies the mock's wire-level contract end-to-end without needing PostgreSQL or ClickHouse running

### DIFF
--- a/test/mock/README.md
+++ b/test/mock/README.md
@@ -1,0 +1,68 @@
+# Misbehaving Mock ClickHouse
+
+Tests that exercise pg_clickhouse against intentionally malformed ClickHouse
+HTTP responses. Companion to issue [#194].
+
+## What it covers
+
+The real ClickHouse server is well-behaved, so the parser (`src/parser.c`)
+and HTTP transport (`src/http.c`, `src/http_streaming.c`) rarely see invalid
+input in CI. PR [#188] showed there are real bugs lurking when the bytes on
+the wire are not the bytes we expect. This mock returns those bytes on
+demand so we can lock the behavior down.
+
+Behaviors covered:
+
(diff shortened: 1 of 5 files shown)


MukundaKatta:
@CLAassistant check

serprex:
idea was more to build around https://aflplus.plus as was being proposed in #171

PR #227, opened 2026-04-28
feat: add three-state secure option (on/off/auto) for TLS control

x4m:
Hi!

Today TLS is chosen over non-obvious heuristics. Maybe let's give a user explicit control over it?

(I'll drop changes to .github/workflows/* after discussion, I needed them for debug)

### DIFF
--- a/.github/workflows/postgres.yml
+++ b/.github/workflows/postgres.yml
@@ -37,8 +37,30 @@ jobs:
         run: cd vendor/clickhouse-cpp && touch -d $(git log -1 --format="@%ct" CMakeLists.txt) CMakeLists.txt
       - name: Test DSO
         run: pg-build-test
+      - name: Show regression diffs (DSO)
+        if: failure()
+        run: cat regression.diffs || true
+      - name: Upload actual results (DSO)
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: regression-results-pg${{ matrix.pg }}-${{ matrix.os.arch }}-dso
+          path: |
+            regression.diffs
+            results/
       - name: Clean
         run: make clean NO_VENDOR_CLEAN=1
       - name: Test Static
         run: pg-build-test
         env: { CH_BUILD: static }
+      - name: Show regression diffs (Static)
+        if: failure()
+        run: cat regression.diffs || true
+      - name: Upload actual results (Static)
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: regression-results-pg${{ matrix.pg }}-${{ matrix.os.arch }}-static
+          path: |
+            regression.diffs
+            results/
--- a/src/include/engine.h
+++ b/src/include/engine.h
@@ -6,6 +6,14 @@
 /*
  * ch_connection_details defines the details for connecting to ClickHouse.
  */
+/* TLS mode for the "secure" FDW option (auto=heuristic, on=force, off=never) */
+typedef enum
+{
+	CH_TLS_AUTO = 0,			/* cloud-hostname heuristic (default) */
+	CH_TLS_ON,					/* always HTTPS; default port 8443 */
+	CH_TLS_OFF,					/* always HTTP;  default port 8123 */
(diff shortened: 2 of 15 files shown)


.github/workflows/postgres.yml
```diff
@@ -37,8 +37,30 @@ jobs:
         run: cd vendor/clickhouse-cpp && touch -d $(git log -1 --format="@%ct" CMakeLists.txt) CMakeLists.txt
       - name: Test DSO
         run: pg-build-test
+      - name: Show regression diffs (DSO)
+        if: failure()
+        run: cat regression.diffs || true
```
theory, line 42: This is not necessary; `pg-build-test` does it.

.github/workflows/postgres.yml
```diff
@@ -37,8 +37,30 @@ jobs:
...
+            regression.diffs
+            results/
       - name: Clean
         run: make clean NO_VENDOR_CLEAN=1
       - name: Test Static
         run: pg-build-test
         env: { CH_BUILD: static }
+      - name: Show regression diffs (Static)
+        if: failure()
+        run: cat regression.diffs || true
```
theory, line 58: Not necessary.

src/include/engine.h
```diff
@@ -6,13 +6,22 @@
...
+}			ch_tls_mode;
+
 typedef struct
 {
 	char	   *host;
 	int			port;
 	char	   *username;
 	char	   *password;
 	char	   *dbname;
+	ch_tls_mode tls;			/* TLS mode; CH_TLS_AUTO when not specified */
```
theory, line 24: No need for the `ch_` prefix here; the struct already has it.

src/include/fdw.h
```diff
@@ -239,7 +239,8 @@ bool		chfdw_pushdown_regex_ok(void);
 
 extern void
 			chfdw_extract_options(List * defelems, char **driver, char **host, int *port,
-								  char **dbname, char **username, char **password);
+								  char **dbname, char **username, char **password,
+								  ch_tls_mode * tls);
```
theory, line 243:
```suggestion
								  tls_mode * tls);
```

src/option.c
```diff
@@ -284,7 +301,8 @@ is_ch_option(const char *keyword)
  */
 void
 chfdw_extract_options(List * defelems, char **driver, char **host, int *port,
-					  char **dbname, char **username, char **password)
+					  char **dbname, char **username, char **password,
+					  ch_tls_mode * tls)
```
theory, line 305:
```suggestion
					  tls_mode * tls)
```

.github/workflows/postgres.yml
```diff
@@ -37,8 +37,30 @@ jobs:
...
+      - name: Upload actual results (DSO)
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
+          name: regression-results-pg${{ matrix.pg }}-${{ matrix.os.arch }}-dso
+          path: |
+            regression.diffs
+            results/
```
theory, line 50: I always look at the logs for result output. What does this achieve?

theory (approved):
Very useful, thank you @x4m. I think we can drop the `ch_` prefix where the other properties or parameters don't use it, and I'm not sure the test uploading stuff is necessary (and the `cat` calls are not), but otherwise it looks great!

theory (changes requested):
Oh, also need to update the binary driver connection stuff in `ch_binary_connect()` and document the new option.

src/include/engine.h
```diff
@@ -6,13 +6,22 @@
 /*
  * ch_connection_details defines the details for connecting to ClickHouse.
  */
+/* TLS mode for the "secure" FDW option (auto=heuristic, on=force, off=never) */
+typedef enum
```
serprex, line 10: this should be able to configure min tls version options too

theory: I think that'd be a separate option. `clickhouse-client` doesn't have an option for this AFAICT.

serprex:
merged in #272

PR #228, opened 2026-04-28
revoke PUBLIC execute on clickhouse_raw_query

x4m:
clickhouse_raw_query(sql, connstring) lets the caller specify an arbitrary host in the connection string. With the function executable by PUBLIC any database user could reach internal services such as cloud metadata endpoints (169.254.169.254), private APIs, or other hosts on the server network directly from the PostgreSQL process (SSRF).

Revoke PUBLIC execute. Administrators who need ad-hoc raw access should grant the function explicitly to a trusted role.


theory:
Wasn't able to push to your branch, so I opened #239 with additional changes. Thank you!

PR #230, opened 2026-05-06
Support mapping to Postgres JSON type

theory:
In addition to the JSONB type mapping, add support for the use of the Postgres JSON type mapped to the ClickHouse JSON type. It already worked for the JSON driver, just hadn't been tested. But adding it to the binary driver required a bit more work.

Mainly that work encompasses making the binary type handling functions aware of the Postgres type used in columns. By default, the binary driver supports a short list of types that map to ClickHouse types, but allows for other types by exercising the functions in `convert.c` to convert between compatible types. These conversion require casts. And while Postgres supports casts between JSON and JSONB, they're `COERCION_PATH_COERCEVIAIO` casts, which `convert.c` doesn't handle.

We could consider adding support for these conversions, but since it requires using the _in and _out functions to do the conversion, and since for JSON all we really need is the text version, this seems unnecessarily wasteful in terms of CPU cycles and memory.

Instead, refactor things so that the binary conversion functions are aware of not only explicitly supported type mappings provided by `get_corr_postgres_type()`, but also the types used for the columns in the foreign tables. Then teach `get_corr_postgres_type()` to inspect these values in the context of the ClickHouse JSON type and allow both.

For fetches (selects), on the first call to `ch_binary_read_row()`, populate `coltypes` with the actual column types and teach `make_datum()` to examine it when considering JSON types. This requires that the `List` of attributes be passed to `ch_binary_read_row`, and also that `coltypes` be initialized with zero values. Ideally `coltypes` would be populated by an earlier hook, before fetching starts, but this will do for now.

For updates (inserts), on the other hand, create a new `List` and populate it with the requisite types in `clickhouseBeginForeignInsert` and `clickhousePlanForeignModify` (not sure why it needs both, but it follows the pattern of setting column attribute numbers, which are also in both places). THen pass it as a new argument to the `prepare_insert` function, now with an updated signature (the HTTP version currently ignores the new argument). Then fetch the types from it in `ch_binary_prepare_insert()` to pass to `get_corr_postgres_type()`.

Teach `init_output_convert_state()` not to try to convert between `JSON` and `JSONB` types, and add comments to explain what's happening with all this type management stuff.

Finally, update the function and operator pushdown functions to support the `->` and `->>` operators for JSON, as well as the `json_extract_path_text()` and `json_extract_path()` functions. Rename the relevant constants to `JSON_*` instead of `JSONB_*`, since they now handle both, and `JSON_*` is the more generic.

Expand the JSON tests to cover all the same patterns for JSON that were previously covered for JSONB, for both the binary and http drivers. This requires additional alternate expected output files due to minor changes to earlier ClickHouse versions that did not support JSON and returned varying error messages about it.

Document the support for JSON, as well as the operators and functions. In fact, the `jsonb_()` functions weren't previously documented, so add them, too. Also fix reversed descriptions for the `->` and `->>` operators and mention that they work for both JSON and JSONB.

While at it, update the exception raised when `column_append()` can't handle a specific Postgres type to emit the name of the unsupported type.

### DIFF
--- a/src/convert.c
+++ b/src/convert.c
@@ -1,3 +1,7 @@
+/*
+ * Conversion functions for the binary driver.
+*/
+
 #include "postgres.h"
 
 #include "funcapi.h"
@@ -188,10 +192,16 @@ ch_binary_convert_datum(void *state, Datum val)
 }
 
 /* input */
+
+/*
+ * Convert val from intype to outtype. No conversion for binary-compatible
+ * types or when intype and outtype are the same. All others converted via the
+ * appropriate CAST function.
+*/
 void	   *
 ch_binary_init_convert_state(Datum val, Oid intype, Oid outtype)
 {
-	/* make_datum() binary.cpp copies all bytes, no cast needed. */
+	/* make_datum() copies all bytes, no cast needed. */
 	if (intype == TEXTOID && outtype == BYTEAOID)
 		return NULL;
 
@@ -360,8 +370,15 @@ init_output_convert_state(ch_convert_output_state * state)
 	if (state->outtype == state->intype)
 		return;
 
-	/* column_append() binary.cpp copies all bytes, no cast needed. */
+	/* column_append() copies all bytes, no cast needed. */
 	if (state->intype == BYTEAOID && state->outtype == TEXTOID)
+		return;
+
+	/* column_append() handles both JSON and JSONB, so no need to convert. */
+	if (
+		(state->intype == JSONBOID && state->outtype == JSONOID)
+		|| (state->intype == JSONOID && state->outtype == JSONBOID)
+		)
 		return;
 
 	/* Postgres has no cast from bool to INT16, so provide our own. */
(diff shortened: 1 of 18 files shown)


src/convert.c
```diff
@@ -188,10 +192,16 @@ ch_binary_convert_datum(void *state, Datum val)
...
+
+/*
+ * Convert val from intype to outtype. No conversion for binary-compatible
+ * types or when intype and outtype are the same. All others converted via the
+ * appropriate CAST function.
+*/
 void	   *
 ch_binary_init_convert_state(Datum val, Oid intype, Oid outtype)
 {
-	/* make_datum() binary.cpp copies all bytes, no cast needed. */
```
serprex, line 194: make_datum was correct, this function is called by `binary_fetch_row`, which gets value with `ch_binary_read_row` which uses `make_datum`

src/convert.c
```diff
@@ -360,10 +370,17 @@ init_output_convert_state(ch_convert_output_state * state)
 	if (state->outtype == state->intype)
 		return;
 
-	/* column_append() binary.cpp copies all bytes, no cast needed. */
+	/* make_datum() copies all bytes, no cast needed. */
```
serprex: `column_append` was correct, this stuff ends up coming from `ch_binary_column_append_data` which uses `column_append`

theory: 🫠

theory: Fixed in 253fdca.

src/convert.c
theory: Fixed in 253fdca.

serprex (approved):

Issue #117, opened 2026-01-06
Add options to control identifier case sensitivity


theory:
Since 54342bb, pg_clickhouse has been using `quote_identifier()` for all tables and columns imported from ClickHouse. This is kind of a pain for mixed case identifiers, because ClickHouse identifiers are case-sensitive, so require no quoting. But mixed case identifiers require quoting in Postgres.

Design an interface to allow lowercase (and therefore unquoted) identifiers to map to mixed case (or uppercase) identifiers in ClickHouse. Perhaps allow a "column_name" attribute, something like:

```
                                    Foreign table "public.hits"
        Column         |            Type             | Collation | Nullable | Default |        FDW options 
-----------------------+-----------------------------+-----------+----------+---------+----------------------------
 watchid               | bigint                      |           | not null |         | (column_name 'WatchID')
 javaenable            | smallint                    |           | not null |         | (column_name 'JavaEnable')
 title                 | text                        |           | not null |         | (column_name 'Title')
 goodevent             | smallint                    |           | not null |         | (column_name 'GoodEvent')
```

This would complement the foreign table `table_name` option, and allow lowercase column names in Postgres to map to mixed-case names in ClickHouse. The deparser would need to read these names and use them in preference to the names provided by Postgres when constructing queries.

Perhaps there could also be an option for [IMPORT FOREIGN SCHEMA](https://www.postgresql.org/docs/current/sql-importforeignschema.html) to tell the importer to prefer lowercase names and only fall back on quoting if there is a conflict with an earlier name.

theory:
Moved the [IMPORT FOREIGN SCHEMA](https://www.postgresql.org/docs/current/sql-importforeignschema.html) bit to #236.

PR #231, opened 2026-05-06
fix column_name option not being respected by inserts

serprex:


test/expected/column_name.out
```diff
@@ -0,0 +1,113 @@
...
+    pg_id int OPTIONS (column_name 'ch_id'),
+    pg_val text OPTIONS (column_name 'ch_val')
+) SERVER cn_bin_loopback OPTIONS (table_name 'td');
+ALTER FOREIGN TABLE cn_drop DROP COLUMN discarded;
+INSERT INTO cn_drop VALUES (7, 'after drop');
+SELECT * FROM cn_drop ORDER BY pg_id;
+ pg_id |   pg_val   
+-------+------------
+     7 | after drop
+(1 row)
```
theory, line 98: Does it work when the column is dropped after an insert and/or select?

theory (approved):
How did I miss the `column_name` option? It needs to be documented in the `CREATE FOREIGN TABLE` section of `pg_clickhouse.md`, just before `AggregateFunction`. Relevant to #117.

Please add mention to `CHANGELOG.md`.

test/expected/column_name.out
serprex: yes

PR #233, opened 2026-05-07
binary: support multidimensional arrays

serprex:
needed for functions like extractAllGroupsHorizontal returning Array(Array(String))

### DIFF
--- a/src/convert.c
+++ b/src/convert.c
@@ -223,7 +223,7 @@ convert_array(ch_convert_state * state, Datum val)
 
 	if (slot->len == 0)
 		val = PointerGetDatum(construct_empty_array(slot->item_type));
-	else if (slot->ndim == 1)
+	else if (slot->ndim <= 1)
 	{
 		void	   *arrout = construct_array(slot->datums, slot->len, slot->item_type,
 											 state->typlen, state->typbyval, state->typalign);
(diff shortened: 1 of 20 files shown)


src/convert.c
```diff
@@ -138,20 +138,160 @@ convert_out_generic(ch_convert_output_state * state, Datum val)
...
+			if (flatten_nested_array(slot, dims, 0, flat, flatnulls, &idx))
+				val = PointerGetDatum(construct_md_array(flat, flatnulls, slot->ndim,
+														 dims, lbs, slot->item_type,
+														 state->typlen, state->typbyval,
+														 state->typalign));
+			else
+			{
+				/*
+				 * Jagged shape: format as text and route through array_in so
+				 * binary surfaces the same malformed-literal error as http.
```
serprex, line 274: this might be overkill, other option is to just raise error that array is jagged

theory: Love consistent error messages!

theory (changes requested):
Great, but I think there are some leftovers, like this in convert.c starting at line 637:

```c
				if (AARR_NDIM(v) > 1)
					ereport(ERROR,
							(errcode(ERRCODE_DATATYPE_MISMATCH),
							 errmsg("pg_clickhouse: inserted array should have one dimension")));
```

And this in `deparse.c` at line 2086:

```c
	if (ndims > 1)
		ereport(ERROR,
				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
				 errmsg("only one dimension of arrays supported by pg_clickhouse")));
```

I presume these are no longer relevant, yes?

Also: Please note improvement in `CHANGELOG.md`.

serprex:
those checks were still relevant because INSERT didn't support multi-dimensional. updated PR to support that

theory (approved):
LFG!

PR #234, opened 2026-05-07
Make TSV parser column-type-aware for bracket-leading strings

iskakaushik:
ClickHouse's TabSeparated format does not escape `[` or `]` in String values, so a value like `[foo]bar` is wire-indistinguishable from an array literal. The HTTP TSV parser previously branched into array-mode purely on a leading `[`, which corrupted any String column whose value started with `[` and aborted with:

```
ERROR: unexpected byte (NN) after array
```

when the byte after `]` wasn't `\t` or `\n`.

Plumb the destination Postgres column's array-ness into `ch_http_read_next` so the parser only enters array-mode when the caller actually expects an array. ClickHouse's own TSV parser is schema-aware for the same reason.

## Test plan

- [x] `make tempcheck` passes locally.
- [x] New regression coverage in `test/sql/http.sql`: bracket-leading String values, streaming-fetch path, and a mixed array+string row.
- [ ] CI matrix across CH 23.3 → 26.3.

### DIFF
--- a/test/expected/http.out
+++ b/test/expected/http.out
@@ -887,6 +887,78 @@ CREATE FOREIGN TABLE ft_bad_fetch (
 ) SERVER http_loopback OPTIONS (table_name 't1', fetch_size '-1');
 ERROR:  invalid value for option "fetch_size": -1
 HINT:  fetch_size must be greater than or equal to 0
+/*
+ * TabSeparated does not escape `[` or `]` in String values, so a value like
+ * `[foo]bar` is wire-indistinguishable from a CH array literal. The parser
+ * uses the destination Postgres column type to decide.
+ */
+SELECT clickhouse_raw_query('CREATE TABLE http_test.t_brk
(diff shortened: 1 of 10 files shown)


test/expected/http.out
```diff
@@ -887,6 +887,78 @@ CREATE FOREIGN TABLE ft_bad_fetch (
...
+ 5  | [a,b,c]_trailing
+ 6  | leading text [bracketed] trailing
+(6 rows)
+
+/*
+ * Mixed row: real Array(String) column alongside a String column whose value
+ * starts with `[`. The parser must stay in sync across columns.
+ */
+SELECT clickhouse_raw_query('CREATE TABLE http_test.t_brk_mix
+    (id String, tags Array(String), term String) ENGINE = MergeTree ORDER BY id');
```
theory, line 941: Seems silly to create more tables when one could just use the original table.

theory (approved):
Looks good. Please also add an item to `CHANGELOG.md`.

PR #235, opened 2026-05-07
Detect TSV NULL marker before unescaping

iskakaushik:
ClickHouse's TabSeparated wire format encodes NULL as the bare 2-byte sequence `\N`, while a literal backslash in data is escaped to `\\`. After parser unescaping, both wire forms `\N` (NULL) and `\\N` (literal `\N` data) collapse to the same 2-char string `\N`, which made the post-parse content-based NULL check in `char_to_datum` misread any `String` value beginning with `\N` as NULL — including a 2-char `\N` that was legitimate data.

Detect the NULL marker in the parser, before unescaping, by recognising the 2-byte `\N` followed by a delimiter or end-of-buffer. Surface it via a new `is_null` flag on `ch_http_read_state` and replace the content check in `char_to_datum` with the flag.

The `gucs` test was inadvertently asserting the buggy behaviour for `format_tsv_null_representation` (whose value is literally `\N`); its expected output is updated to reflect the correct round-trip.

## Repro (pre-fix)

```sql
-- ClickHouse:
INSERT INTO t (s) VALUES ('\\N foo bar');  -- 10-char literal data

-- Postgres via pg_clickhouse (HTTP/TSV):
SELECT s, s IS NULL FROM ft;
--  s | ?column?
-- ---+----------
--    | t           <- WRONG: should return '\N foo bar', not NULL
```

## Test plan

- [x] `make tempcheck` against ClickHouse 25.x — all 28 tests pass.
- [x] New regression coverage in `test/sql/http.sql`: literal `\N` data values (exact 2-char, prefix, mid-string, plain).
- [ ] CI matrix across CH 23.3 → 26.3.

## Note

Independent of #234 (bracket-leading `String` parser fix). Both PRs branch off `main`. If #234 lands first, this PR's `test/sql/http.sql` and `test/expected/http*.out` will need a small rebase to insert the `t_nullmark` section in a different anchor.


test/sql/gucs.sql
```diff
@@ -136,7 +136,8 @@ SELECT remote.name, remote.value IS NOT DISTINCT FROM def.value
 WHERE remote.name = ANY(:'set_list')
 ORDER BY remote.name;
 
--- date_time_output_format and format_tsv_null_representation are distinct.
```
theory, line 139: `format_tsv_null_representation` is still distinct. Why remove it from this comment?

theory (approved):
Thank you. One question, otherwise just need a `CHANGELOG.md` entry.

test/sql/gucs.sql
theory: Restored it with a bit more detail in ebb78e28.

PR #244, opened 2026-05-09
Push down compatible to_char

serprex:
Translates PG to_char(timestamp[tz], fmt) to CH formatDateTime when fmt is a constant whose every keyword has CH equivalent: YYYY/YY, MM, DD, DDD, HH24, HH12/HH, MI, SS, Q, Mon, Dy, AM/PM, plus lowercase variants

### DIFF
--- a/doc/pg_clickhouse.md
+++ b/doc/pg_clickhouse.md
@@ -1087,6 +1087,11 @@ maps the following functions:
 *   `jsonb_extract_path_text`: [sub-column syntax](https://clickhouse.com/docs/sql-reference/data-types/newjson#reading-json-paths-as-sub-columns)
 *   `jsonb_extract_path`: [toJSONString](https://clickhouse.com/docs/sql-reference/functions/json-functions#toJSONString) + [sub-column syntax](https://clickhouse.com/docs/sql-reference/data-types/newjson#reading-json-paths-as-sub-columns)
 *   `to_timestamp(float8)`: [fromUnixTimestamp](https://clickhouse.com/docs/sql-reference/functions/date-time-functions#fromUnixTimestamp)
+*   `to_char(timestamp[tz], fmt)`: [formatDateTime](https://clickhouse.com/docs/sql-reference/functions/date-time-functions#formatDateTime)
+    when `fmt` is a string constant whose every keyword has a faithful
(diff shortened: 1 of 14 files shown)


src/deparse.c
```diff
@@ -2546,6 +2546,172 @@ appendRegex(List * args, deparse_expr_cxt * context)
...
+						appendStringInfoString(out, "%%");
+					else
+						appendStringInfoChar(out, *p);
+				}
+				p++;
+			}
+			continue;
+		}
+
+		/* Outside quotes, backslash escapes only a literal " */
```
theory, line 2674: And not backslash itself? Interesting!

theory (changes requested):
Very nice, thank you. Please document it; would be nice to document exactly what translates and what doesn't.

serprex:
updated with docs

src/deparse.c
serprex: https://github.com/postgres/postgres/commit/63ca86318dc3d6a768eed78efbc6ca014a0622a8

doc/pg_clickhouse.md
```diff
@@ -1319,6 +1324,45 @@ Postgres regular expression from pushing down to ClickHouse, and using the
...
+| `DD`, `dd` | `%d` | zero-padded day of month (01–31) |
+| `DDD`, `ddd` | `%j` | zero-padded day of year (001–366) |
+| `HH24`, `hh24` | `%H` | zero-padded 24-hour (00–23) |
+| `HH`, `hh`, `HH12`, `hh12` | `%I` | zero-padded 12-hour (01–12) |
+| `MI`, `mi` | `%i` | zero-padded minute (00–59) |
+| `SS`, `ss` | `%S` | zero-padded second (00–59) |
+| `Q`, `q` | `%Q` | quarter (1–4) |
+| `Mon` | `%b` | abbreviated month name, e.g., `Oct` |
+| `Dy` | `%a` | abbreviated weekday name, e.g., `Mon` |
+| `AM`, `PM` | `%p` | meridiem indicator, always uppercase |
```
theory, line 1361: Wow, is that all?

theory (approved):

PR #245, opened 2026-05-09
Don't push down functions by default

serprex:
Explicitly add functions which were falling through correctly by default

Include mapping mod/pow/power/bit_count/reverse/bit_and/bit_or/bit_xor

`length(text)` in PG is actually `lengthUTF8` in CH

asin/acos/atanh/acosh not pushed down for now since they need range checks: on PG out-of-range is error, whereas on CH out-of-range returns NaN

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ ALTER EXTENSION pg_clickhouse UPDATE TO '0.3';
     ClickHouse equivalents (e.g., `re2match` → `match`, `re2extractall` →
     `extractAll`). Thanks to Philip Dubé for the PR ([#204]).
 *   Added pushdown for [fuzzystrmatch] functions `soundex()` and
-    `levenshtein()` (2-arg, mapped to `editDistance`).  Thanks to
+    `levenshtein()` (2-arg, mapped to `editDistanceUTF8`). Thanks to
     Philip Dubé for the PR ([#210]).
 *   Added mapping for `JSON` => `jsonb` to the binary driver (requires
     ClickHouse 24.10 or later).
@@ -37,6 +37,12 @@ ALTER EXTENSION pg_clickhouse UPDATE TO '0.3';
     identical CH equivalent (`YYYY`, `MM`, `DD`, `DDD`, `HH24`, `HH12`, `HH`,
     `MI`, `SS`, `Q`, `Mon`, `Dy`, `AM`/`PM`, plus lowercase variants).
     Thanks to Philip Dubé for the PR ([#244]).
+*   Made builtin function pushdown opt-in: Postgres builtins now ship to
+    ClickHouse only when explicitly mapped, so name or signature differences
+    cannot silently alter results. Thanks to Philip Dubé for the PR ([#245]).
+*   Added explicit mappings for `mod`, `pow`/`power`, `bit_count(bytea)`, and
+    `reverse(text)` (→ `reverseUTF8`) to retain previously working pushdowns.
+    Thanks to Philip Dubé for the PR ([#245]).
 
 ### 🐞 Bug Fixes
 
@@ -53,6 +59,14 @@ ALTER EXTENSION pg_clickhouse UPDATE TO '0.3';
     `INSERT`, which caused the binary engine to fail to match ClickHouse block
     columns and the HTTP engine to deparse PostgreSQL attribute names. Thanks
     to Philip Dubé for the PR ([#231]).
+*   Fixed `length(text)` and `strpos(text, text)` pushdown to map to
+    `lengthUTF8` and `positionUTF8` rather than ClickHouse's byte-counting
+    `length` and `position`, matching Postgres character semantics. Thanks to
+    Philip Dubé for the PR ([#245]).
+*   Stopped pushing down `asin`, `acos`, `atanh`, and `acosh`: Postgres raises
+    an error on out-of-range input where ClickHouse returns `NaN`. Evaluating
+    locally preserves Postgres semantics. Thanks to Philip Dubé for the PR
+    ([#245]).
 
 ### 📚 Documentation
 
@@ -113,6 +127,8 @@ ALTER EXTENSION pg_clickhouse UPDATE TO '0.3';
     "Wikipedia: Server-side request forgery"
   [#228]: https://github.com/ClickHouse/pg_clickhouse/pull/228
     "pg_clickhouse#228 Security: revoke PUBLIC execute on clickhouse_raw_query"
+  [#245]: https://github.com/ClickHouse/pg_clickhouse/pull/245
+    "ClickHouse/pg_clickhouse#245 Don't push down functions by default"
 
 ## [v0.2.0] — 2026-04-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file. It uses the
   [Semantic Versioning]: https://semver.org/spec/v2.0.0.html
     "Semantic Versioning 2.0.0"
 
-## [v0.3.0] — Unreleased
+## [v0.3.0] — 2026-05-11
 
 This release makes binary-compatible changes to the v0.2 releases. Once
 installed, any existing use of pg_clickhouse v0.2 will benefit from its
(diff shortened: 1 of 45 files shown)


src/custom_types.c
```diff
@@ -365,9 +631,32 @@ chfdw_check_for_custom_function(Oid funcid)
 				}
 			case F_STRPOS:
 				{
-					strcpy(entry->custom_name, "position");
+					/* PG strpos counts code points, CH position counts bytes */
+					strcpy(entry->custom_name, "positionUTF8");
```
theory, line 641: Hrm. Might we want to vary this depending on the Postgres character encoding? We've kind of been ignoring encodings…

src/custom_types.c
```diff
@@ -408,6 +697,54 @@ chfdw_check_for_custom_function(Oid funcid)
 					entry->paren_count = 3;
 					break;
 				}
+			case F_REVERSE_TEXT:
+				/* reverse code points, not bytes */
```
theory, line 707: But one can store bytes in CH String…

doc/pg_clickhouse.md
```diff
@@ -1033,12 +1033,18 @@ SELECT clickhouse_raw_query(
...
 ### Pushdown Functions
 
-All PostgreSQL builtin functions used in conditionals (`HAVING` and `WHERE`
-clauses) to query ClickHouse foreign tables automatically push down to
-ClickHouse with the same names and signatures. However, some have different
-names or signatures and must be mapped to their equivalents. `pg_clickhouse`
-maps the following functions:
-
+`pg_clickhouse` only pushes down the PostgreSQL builtin functions listed
+below.
```
```suggestion
pg-clickhouse pushes down a subset of the PostgreSQL builtin functions used in
conditionals (`HAVING` and `WHERE` clauses). That subset map to ClickHouse
equivalents as follows:
```

CHANGELOG.md
```diff
@@ -43,6 +49,12 @@ ALTER EXTENSION pg_clickhouse UPDATE TO '0.3';
     `INSERT`, which caused the binary engine to fail to match ClickHouse block
     columns and the HTTP engine to deparse PostgreSQL attribute names. Thanks
     to Philip Dubé for the PR ([#231]).
+*   Fixed `length(text)` pushdown to map to `lengthUTF8` rather than
```
theory: Also `position()`, no?

test/expected/functions.out
```diff
@@ -1769,6 +1769,99 @@ SELECT * FROM t4 WHERE levenshtein(val, 'val1', 1, 1, 2) <= 1;
...
+
+SELECT val FROM t4 WHERE reverse(val) = 'βαΩ';
+ val 
+-----
+ Ωαβ
+(1 row)
+
+-- bit_count(bytea) pushes down as bitCount (PG14+).
+SELECT current_setting('server_version_num')::int >= 140000 AS pg14 \gset
+\if :pg14
```
theory, line 1929: Yes! I've been thinking of using psql conditionals instead of PL/pgSQL. Much nicer.

doc/pg_clickhouse.md
```diff
@@ -1033,12 +1033,18 @@ SELECT clickhouse_raw_query(
...
+
+*   `abs`: [abs](https://clickhouse.com/docs/sql-reference/functions/arithmetic-functions#abs)
+*   `factorial`: [factorial](https://clickhouse.com/docs/sql-reference/functions/math-functions#factorial)
+*   `mod` (int2/int4/int8/numeric): [modulo](https://clickhouse.com/docs/sql-reference/functions/arithmetic-functions#modulo)
+*   `pow` & `power` (float8/numeric): [pow](https://clickhouse.com/docs/sql-reference/functions/math-functions#pow)
+*   `round`: [round](https://clickhouse.com/docs/sql-reference/functions/rounding-functions#round)
+*   `sin`, `cos`, `tan`, `atan`, `atan2`, `sinh`, `cosh`, `tanh`, `asinh`,
+    `degrees`, `radians`, `pi`: [ClickHouse math functions](https://clickhouse.com/docs/sql-reference/functions/math-functions)
+    of the same name. `asin`, `acos`, `atanh`, `acosh` are not pushed
+    down: PG raises on out-of-range input where CH returns `NaN`.
```
theory, line 1048: Shouldn't we add tests for these?

doc/pg_clickhouse.md
```diff
@@ -1077,11 +1086,23 @@ maps the following functions:
 *   `array_sample`: [arrayRandomSample](https://clickhouse.com/docs/sql-reference/functions/array-functions#arrayRandomSample)
 *   `array_sort`: [arraySort](https://clickhouse.com/docs/sql-reference/functions/array-functions#arraySort) / [arrayReverseSort](https://clickhouse.com/docs/sql-reference/functions/array-functions#arrayReverseSort)
 *   `btrim`: [trimBoth](https://clickhouse.com/docs/sql-reference/functions/string-functions#trimboth)
-*   `strpos`: [position](https://clickhouse.com/docs/sql-reference/functions/string-search-functions#position)
+*   `ltrim`: [ltrim](https://clickhouse.com/docs/sql-reference/functions/string-functions#ltrim)
+*   `rtrim`: [rtrim](https://clickhouse.com/docs/sql-reference/functions/string-functions#rtrim)
+*   `concat_ws`: [concatWithSeparator](https://clickhouse.com/docs/sql-reference/functions/string-functions#concatwithseparator)
+*   `lower(text)`: [lowerUTF8](https://clickhouse.com/docs/sql-reference/functions/string-functions#lowerutf8)
+*   `upper(text)`: [upperUTF8](https://clickhouse.com/docs/sql-reference/functions/string-functions#upperutf8)
```
theory, line 1094: No test for this one?

doc/pg_clickhouse.md
```diff
@@ -1182,12 +1203,16 @@ These PostgreSQL aggregate functions pushdown to ClickHouse.
 
 *   [array_agg](https://clickhouse.com/docs/sql-reference/aggregate-functions/reference/grouparray)
 *   [avg](https://clickhouse.com/docs/sql-reference/aggregate-functions/reference/avg)
+*   [bit_and](https://clickhouse.com/docs/sql-reference/aggregate-functions/reference/groupbitand)
+*   [bit_or](https://clickhouse.com/docs/sql-reference/aggregate-functions/reference/groupbitor)
+*   [bit_xor](https://clickhouse.com/docs/sql-reference/aggregate-functions/reference/groupbitxor)
```
theory, line 1218: Need tests?

theory (changes requested):
Much better than before. Good to have this cleaned up to avoid the legacy of the fork from postgres_fdw.

Seems like we could use a few more tests though, yes?

src/custom_types.c
serprex: outside ascii/utf8 we wouldn't be able to push anything down. would prefer to keep that out of scope for now

src/custom_types.c
serprex: for that PG18 added `reverse(bytea)`

serprex:
heavily expanded test suite. also fixed deparsing bytea, as ch doesn't support pg bytea syntax for strings

test/expected/aggregates.out
```diff
@@ -1334,6 +1334,42 @@ FDW options: (database 'agg_test', table_name 'hits', engine 'MergeTree')
  t
 (1 row)
 
+-- bit_and pushdown (binary)
+            QUERY PLAN            
+----------------------------------
+ Foreign Scan
+   Relations: Aggregate on (hits)
+(2 rows)
```
theory: Could we have `VERBOSE` to see the remote query?

test/expected/window_functions.out
```diff
@@ -115,6 +115,60 @@
  lead_300  | 2026-03-05 08:00:00+00 | 1970-01-01 00:00:00+00
 (5 rows)
 
+-- LAG pushdown (binary)
+           QUERY PLAN            
+---------------------------------
+ Foreign Scan
+   Relations: Window on (events)
```
theory, line 122: Why no verbose?

theory (approved):
Great! I can add the missing `VERBOSE`s and merge.

src/custom_types.c
theory: Yeah, I realized it's TEXT, so needs to be strings.

test/expected/aggregates.out
theory: Done in 6a0cb8e.

test/expected/window_functions.out
theory: Fixed in 6a0cb8e.

PR #246, opened 2026-05-09
Do not run tests twice when PR submitted

serprex:
https://github.com/wal-g/wal-g/pull/1931

### DIFF
--- a/.github/workflows/clickhouse.yml
+++ b/.github/workflows/clickhouse.yml
@@ -1,6 +1,7 @@
 name: 🏠 ClickHouse CI
 on:
   push:
+    branches: [main, wip/**]
   pull_request:
   schedule:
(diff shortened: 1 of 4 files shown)


theory (commented):
This will prevent tests from running when there is no PR. though. I often have WIP branches that I may or may not want tests to run on before creating a PR.

serprex:
> This will prevent tests from running when there is no PR. though. I often have WIP branches that I may or may not want tests to run on before creating a PR.

I usually use drafts for that, but fair enough

theory:
What I've done on other projects is suppress tests in a branch starting with `wip/`. Here, perhaps it makes sense to allow them with some other prefix.

theory:
Maybe

```
    branches: [main, wip/**]
```

theory (approved):

PR #254, opened 2026-05-21
Replace clickhouse-cpp with clickhouse-c

### LINKED ISSUE #241: Audit and fix binary driver RAII vs PG_TRY setjmp/longjmp issues
Review `binary.cpp` for conflicts between Postgres's `PG_TRY` setjmp/longjmp stuff and C++'s [RAII](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization). Remove as many conflicts a possible, and document with detailed comments those that cannot be fixed.

Initial list of issues from @serprex:

- make_datum can cause PG exception, breaking RAII. eg cstring_to_text_with_len
- ch_binary_simple_query exception handler can dereference nullptr
- ResetConnection is called in exception handlers, which can rethrow an uncaught exception
- elog(ERROR) in catch block isn't safe: https://docs.oracle.com/cd/E19205-01/819-5267/bkahg/index.html
- CreateTemplateTupleDesc can OOM & leak block
- elog(ERROR) in ch_binary_insert_state_free is questionable, kind of like raising an exception in a c++ destructor
- ch_binary_close doesn't free conn
- error being set with strdup, not checking if that returns null

serprex:
Resolves #241 and #254.


README.md
```diff
@@ -95,7 +94,6 @@ sudo yum install \
   libuuid-devel \
   openssl-libs \
   automake \
-  cmake \
```
theory, line 98: Need to add `liblz4` and `libzstd1`, yes?

README.md
```diff
@@ -239,8 +225,8 @@ CREATE EXTENSION pg_clickhouse SCHEMA env;
 ## Dependencies
 
 The `pg_clickhouse` extension requires [PostgreSQL] 13 or higher, [libcurl],
-[libuuid]. Building the extension requires a C and C++ compiler, [libSSL], [GNU
-make], and [CMake].
+[libuuid]. Building the extension requires a C compiler, [libSSL], and [GNU
+make].
```
theory: Also liblz4 and libzstd1, yes?

src/binary/insert.c
```diff
@@ -0,0 +1,2243 @@
...
+	if (!info)
+		return false;
+	if (info->version_major > 24)
+		return true;
+	if (info->version_major == 24 && info->version_minor >= 10)
+		return true;
+	return false;
+}
+
+/* dynamic buffer for raw bytes; palloc-backed, freed by context delete */
```
serprex, line 24: could use StringInfoData if we're okay choking on blocks larger than 1GB

src/binary.c
```diff
@@ -0,0 +1,2243 @@
...
+			 errmsg("pg_clickhouse: %s%s", prefix ? prefix : "", m)));
+}
+
+/* ===================================================================== */
+/* Connection                                                             */
+/* ===================================================================== */
+
+struct ch_binary_state
+{
+	MemoryContext cxt;
```
serprex: @theory I know you've pointed out that this sort of thing belongs in a library, but hooking into MemoryContext / ereport / etc will hopefully convince you of the cleanliness keeping this out of `clickhouse-c`. maybe we need a `src/binary/` with binary/binary_decode/binary_encode/binary_connection/select/insert/...

src/binary.c
```diff
@@ -0,0 +1,2243 @@
...
+		const		uint8_t *bytes = c->body.data + start;
+
+		/* If nullable and this row was null (signalled by null bit), key 0. */
+		if (nullable && c->nulls.data[i])
+		{
+			keys[i] = 0;
+			continue;
+		}
+
+		/* Linear search current dict (skip slot 0 for Nullable). */
```
serprex: this might be trusting user to understand what "low cardinality" means too much

src/binary.c
theory: I'm not saying it "belongs in a library" so much as "it sure would be nice if this stuff was also offered in a library". :-) I get the challenge with integrating Postgres features like `palloc` and `ereport`; I could also see those being callbacks.

src/binary.c
theory: Well I understand "low cardinality" but don't see that phrase in this comment. Maybe make the comment more explanatory?

serprex: seeing as https://github.com/ClickHouse/pg_stat_ch/pull/99 is going with 270 cardinality `LowCardinality(String)` I'm thinking this should use something better than linear search anyways, will tidy then. docs explicitly recommend with up to 1000 values

src/binary/binary.c
```diff
@@ -0,0 +1,105 @@
+/*
+ * binary.c
+ *
+ * Core glue for the binary driver. Owns the chc_alloc thunks (palloc on
+ * CurrentMemoryContext, MCXT_ALLOC_HUGE so block buffers escape the 1GB
+ * cap), error mapping into ereport, and a couple of small type helpers
+ * shared across the subdir.
+ *
+ * This TU also holds the CHC_IMPLEMENTATION define for the clickhouse-c
```
theory: What does "TU" stand for?

src/binary/binary.c
```diff
@@ -0,0 +1,105 @@
...
+	if (chc_type_kind(t) == CHC_NULLABLE)
+		t = chc_type_child(t, 0);
+	if (chc_type_kind(t) == CHC_LOW_CARDINALITY)
+	{
+		t = chc_type_child(t, 0);
+		if (chc_type_kind(t) == CHC_NULLABLE)
+			t = chc_type_child(t, 0);
+	}
+	return t;
+}
```
theory: What is the purpose of this function? I can suss out the others, but could use some documentation for this one.

src/binary/connection.c
```diff
@@ -0,0 +1,259 @@
...
+		SSL_load_error_strings();
+		OpenSSL_add_all_algorithms();
+		openssl_init_done = 1;
+	}
+
+	s->ssl_ctx = SSL_CTX_new(TLS_client_method());
+	if (!s->ssl_ctx)
+		ereport(ERROR,
+				(errcode(ERRCODE_FDW_UNABLE_TO_ESTABLISH_CONNECTION),
+				 errmsg("pg_clickhouse: SSL_CTX_new failed")));
```
theory: Is there a more user-friendly error we can return? User may not know what `SSL_CTX_new` is (I don't). Same for the `SSL_new` failure below.

src/binary/connection.c
```diff
@@ -0,0 +1,259 @@
...
+		}
+		else
+			port = 9000;
+	}
+	else if (port == CLICKHOUSE_SECURE_PORT)
+		tls = true;
+
+	MemoryContext cxt = AllocSetContextCreate(CacheMemoryContext,
+											  "pg_clickhouse binary connection",
+											  ALLOCSET_SMALL_SIZES);
```
theory, line 187: So is the idea here that the binary client gets its own memory context? I presume it's used for fetching results, yes? Maybe some notes for the `ch_binary_state` fields in `binary_internal.h` would be useful.

src/binary/binary.c
serprex: translation unit, will omit here as it's clear enough from context

src/binary/connection.c
serprex: would `failure to initialize openssl` work? with errdetail saying the exact openssl call that failed?

src/binary/binary.c
serprex: it unwraps Nullable/LowCardinality so that callers can do type dispatch on the inner type. inlined

src/binary/connection.c
serprex: `CacheMemoryContext` is a long lived context we use to keep connection between transactions

src/binary/connection.c
theory: I might lean away from citing the name of the library, but otherwise yes, that seems better to me.

src/binary/encode.c
```diff
@@ -147,15 +147,6 @@ ch_binary_prepare_insert(void *conn, const ch_query * query,
 	}
 }
 
-/* Any CH integer kind whose value fits in int64 (column width truncates). */
```
theory: I liked having this explanation.

src/binary/decode.c
```diff
@@ -0,0 +1,900 @@
...
+	uint64_t	end = offs[row];
+
+	*out_ptr = (const char *) data + start;
+	*out_len = (size_t) (end - start);
+}
+
+static Oid
+ch_kind_to_pg_oid(const chc_type * type)
+{
+	switch (chc_type_kind(type))
```
theory, line 139: No time or time64?

src/binary/decode.c
```diff
@@ -0,0 +1,900 @@
...
+format_decimal_text(const uint8_t * bytes, size_t width, uint32_t scale,
+					char *out, size_t out_cap)
+{
+	uint8_t		mag[32];
+	int			neg = 0;
+
+	if (width > 32 || width == 0)
+		return -1;
+	memcpy(mag, bytes, width);
+	if (mag[width - 1] & 0x80)
```
theory: What does this mean about `mag`? Honestly wouldn't mind some comments to explain what all's going on in this function. It looks nothing like the [old C++ version](https://github.com/ClickHouse/pg_clickhouse/blob/a84662aa3ae7039ba09621183600e7b0e2673988/src/binary.cpp#L1041-L1087).

src/binary/decode.c
```diff
@@ -0,0 +1,900 @@
...
+
+	slot->len = len;
+	slot->ndim = ndim;
+	slot->item_type = ch_kind_to_pg_oid(leaf);
+	slot->array_type = get_array_type(slot->item_type);
+	if (slot->array_type == InvalidOid)
+		ereport(ERROR,
+				(errcode(ERRCODE_FDW_INVALID_DATA_TYPE),
+				 errmsg("pg_clickhouse: could not find array type for column type %u",
+						slot->item_type)));
```
theory: Can we show the name of the type here?

src/binary/decode.c
```diff
@@ -0,0 +1,900 @@
...
+		case CHC_INT8:
+			*valtype = INT2OID;
+			return (Datum) (int16) rd_i8(fixed_data(col), row);
+		case CHC_INT16:
+			*valtype = INT2OID;
+			return (Datum) rd_i16(fixed_data(col), row);
+		case CHC_UINT16:
+
+			/*
+			 * Legacy binary FDW reads UInt16 via narrowing to int16, so 65535
```
theory: Oh not int32? Seems like a bug, no?

src/binary/decode.c
```diff
@@ -0,0 +1,900 @@
...
+		case CHC_ENUM16:
+			*valtype = TEXTOID;
+			return read_enum_as_text(col, type, row);
+		case CHC_JSON:
+		case CHC_OBJECT:
+			{
+				/*
+				 * *valtype arrives set to JSONBOID by default (from
+				 * ch_kind_to_pg_oid) but the caller's foreign-table
+				 * declaration may have requested JSONOID; honour it.
```
theory: This was a change I made intentionally. Seems as if there might be other situations in which we want to use the declared type, no? Otherwise, it’s later using casting, which seems a little weird. Not that we want to fix it here. Just to say that it looks like a special case here, but I’m thinking it could become more normal.

src/binary/decode.c
```diff
@@ -0,0 +1,900 @@
...
+		case CHC_IPV6:
+			*valtype = INETOID;
+			return read_ipv6(col, row);
+		case CHC_LOW_CARDINALITY:
+			return read_lc_string(col, type, row, valtype, is_null);
+		case CHC_ARRAY:
+			return read_array(col, type, row, valtype, is_null);
+		case CHC_TUPLE:
+			return read_tuple(col, type, row, valtype, is_null);
+		default:
```
theory, line 739: Very much appreciating the cleanliness of the functional dispatch in this switch statement.

src/binary/decode.c
```diff
@@ -0,0 +1,900 @@
...
+read_decimal(const chc_column * col, const chc_type * type, uint64_t row)
+{
+	size_t		es;
+	const		uint8_t *p = (const uint8_t *) chc_column_fixed_data(col, &es);
+	uint32_t	scale = (uint32_t) chc_type_decimal_scale(type);
+	char		buf[80];
+	int			rc;
+
+#if PG_VERSION_NUM >= 140000
+	/* Decimal32/64 fit in int64; skip the byte-array text path. */
```
theory, line 293: Oooh, I was not aware of these functions! Very nice find.

src/binary/encode.c
```diff
@@ -0,0 +1,408 @@
...
+#include "utils/date.h"
+#include "utils/inet.h"
+#include "utils/lsyscache.h"
+#include "utils/palloc.h"
+#include "utils/timestamp.h"
+#include "utils/uuid.h"
+
+#include "binary.h"
+
+/* power-of-10 lookup; CH bounds DateTime64 scale to [0, 9] */
```
theory: Dupe? Move to internal header?

src/binary/encode.c
```diff
@@ -0,0 +1,408 @@
...
+			}
+		case CHC_TUPLE:
+			return RECORDOID;
+		case CHC_NULLABLE:
+		case CHC_LOW_CARDINALITY:
+			return ch_kind_to_pg_oid_for_insert(chc_type_child(type, 0), colname);
+		default:
+			ereport(ERROR,
+					(errcode(ERRCODE_FDW_INVALID_DATA_TYPE),
+					 errmsg("pg_clickhouse: unsupported column type for \"%s\"",
```
theory: Can we show the unsupported type name in this error message?

src/binary/encode.c
```diff
@@ -0,0 +1,408 @@
...
+				if (!isnull)
+					seconds = ((int64_t) DatumGetDateADT(val)
+							   + CH_TO_PG_DATE_OFFSET) * SECS_PER_DAY;
+				ch_binary_append_date_seconds(h, colidx, seconds, isnull);
+				return;
+			}
+		case TIMESTAMPOID:
+		case TIMESTAMPTZOID:
+			{
+				if (kind == CHC_DATETIME)
```
theory: Would a switch statement be slightly cleaner looking here? Nitpicking, sorry.

src/binary/insert.c
```diff
@@ -0,0 +1,1477 @@
...
+ * Per-column accumulator. Layout decided at begin_insert from the column's
+ * chc_type. Each typed append_* appends into one of three storage groups:
+ *   - body / body_offs:  fixed-width or string values for the top-level
+ *     column or, when an array context is open, the inner items.
+ *   - arr_offs:          cumulative outer ends for Array(*) columns.
+ *   - nulls:             one byte per top-level row for Nullable(*).
+ */
+typedef enum
+{
+	IC_FIXED,
```
theory, line 110: IC == insert ClickHouse ?

src/binary/insert.c
```diff
@@ -0,0 +1,1477 @@
...
+			const		chc_type *ct = chc_block_column_type(h->initial_block, i);
+			size_t		nlen;
+			const char *nm;
+
+			classify_column(c, ct);
+			nm = chc_block_column_name(h->initial_block, i, &nlen);
+			c->info.name = pnstrdup(nm ? nm : "", nlen);
+			c->info.is_nullable = c->is_nullable;
+
+			/* inner_t already Nullable-stripped; unwrap LowCardinality
```
theory: What does “Nullable-stripped” mean?

src/binary/insert.c
```diff
@@ -0,0 +1,1477 @@
...
+	ic_col	   *c = h->array_active ? &h->cols[h->array_col_idx] : &h->cols[col_idx];
+
+	if (isnull && !h->array_active && !c->is_nullable)
+	{
+		size_t		tnlen;
+		const char *tname = chc_type_name(c->t, &tnlen);
+
+		ereport(ERROR,
+				(errcode(ERRCODE_NOT_NULL_VIOLATION),
+				 errmsg("pg_clickhouse: could not append data to column - cannot append NULL to NOT NULL %.*s column",
```
theory: This error message seems repetitive to me

src/binary/insert.c
```diff
@@ -0,0 +1,1477 @@
...
+
+		if (!isnull && addr_be)
+			memcpy(raw, addr_be, 16);
+		append_fixed_bytes(c, raw, 16);
+		MemoryContextSwitchTo(old);
+		return;
+	}
+	ereport(ERROR,
+			(errcode(ERRCODE_DATATYPE_MISMATCH),
+			 errmsg("pg_clickhouse: inet into non-inet column")));
```
theory: Should the error message start with “cannot map” or something like that?

src/binary/binary.c
```diff
@@ -0,0 +1,89 @@
...
+ * connection.c / select.c / insert.c live in binary_internal.h
+ * alongside this file.
+ */
+
+#include "postgres.h"
+
+#include "utils/memutils.h"
+#include "utils/palloc.h"
+
+#define CHC_IMPLEMENTATION
```
theory, line 23: Should this move to the header file? Now that the files are split up, VSCode, at least, doesn't resolve symbols defined when this is defined in any file but `binary.c` Hrm, tried that, didn't help. :-(

src/binary/decode.c
```diff
@@ -0,0 +1,900 @@
...
+					 errmsg("pg_clickhouse: unsupported type in binary protocol")));
+	}
+	/* unreachable */
+	return (Datum) 0;
+}
+
+/* ---- read state ----------------------------------------------------- */
+
+static bool
+load_block(ch_binary_read_state_t * state)
```
theory, line 751: If it's just loading one block at a time, is it streaming block by block or reading them all into memory elsewhere?

src/binary/decode.c
```diff
@@ -0,0 +1,900 @@
...
+		state->cur = NULL;
+		goto again;
+	}
+
+	PG_TRY();
+	{
+		for (size_t i = 0; i < ncols; i++)
+		{
+			/*
+			 * For most types read_value overwrites *valtype with the
```
theory:
```suggestion
			 * Currently, read_value overwrites *valtype for most types with the
```

src/binary/insert.c
```diff
@@ -0,0 +1,1477 @@
...
+#include <string.h>
+
+#include "common/hashfn.h"
+#include "port/pg_bswap.h"
+#include "utils/memutils.h"
+#include "utils/palloc.h"
+
+#include "binary_internal.h"
+
+/* dynamic buffer for raw bytes; palloc-backed, freed by context delete */
```
theory, line 24: Him, does PostgreSQL not have a data structure like this? I guess `StringInfoData` is close, but guarantees null termination, which presumably we don't want.

src/binary/insert.c
```diff
@@ -0,0 +1,1477 @@
...
+	if (k == CHC_LOW_CARDINALITY)
+	{
+		const		chc_type *inner = chc_type_child(t, 0);
+		bool		inner_nullable = chc_type_kind(inner) == CHC_NULLABLE;
+		const		chc_type *base = inner_nullable ? chc_type_child(inner, 0) : inner;
+
+		if (chc_type_kind(base) != CHC_STRING)
+			ereport(ERROR,
+					(errcode(ERRCODE_FDW_INVALID_DATA_TYPE),
+					 errmsg("pg_clickhouse: unsupported LowCardinality variant")));
```
theory: Can the error message include the name of the type that's not supported?

src/binary/insert.c
```diff
@@ -0,0 +1,1477 @@
...
+		int			ndim = 0;
+
+		while (chc_type_kind(base) == CHC_ARRAY)
+		{
+			ndim++;
+			base = chc_type_child(base, 0);
+		}
+		bool		elem_nullable = chc_type_kind(base) == CHC_NULLABLE;
+
+		(void) elem_nullable;	/* not supported yet */
```
theory: What's happening here? We support nullable columns, no?

src/binary/insert.c
```diff
@@ -0,0 +1,1477 @@
...
+		{
+			ic->layout = ndim == 1 ? IC_ARRAY_STRING : IC_ARRAY_NESTED_STRING;
+			ic->array_inner_is_string = true;
+		}
+		else if (chc_type_elem_size(base) > 0)
+		{
+			ic->layout = ndim == 1 ? IC_ARRAY_FIXED : IC_ARRAY_NESTED_FIXED;
+			ic->elem_size = chc_type_elem_size(base);
+		}
+		else
```
theory, line 248: IIUC we reach this block if the base type isn't a string and isn't of fixed size. Does that mean something like JSON isn't supported in arrays? (I mean it's silly, but surely there are other varlena types…)

src/binary/insert.c
```diff
@@ -0,0 +1,1477 @@
...
+		dynbuf_append(&c->body, p, n);
+	u64buf_push(&c->body_offs, c->body.len);
+}
+
+/*
+ * Convert decimal text "[-]digits[.frac]" to little-endian bytes of width
+ * `width` honouring `scale`.
+ */
+static void
+decimal_text_to_bytes(const char *s, uint32_t scale, size_t width, uint8_t * out)
```
theory, line 546: Would like comments explaining what this thing is doing in each step.

src/binary/select.c
```diff
@@ -0,0 +1,408 @@
...
+		return;
+
+	MemoryContext old = MemoryContextSwitchTo(resp->cxt);
+	chc_err		ce = {0};
+
+	(void) chc_client_send_cancel(resp->client, &ce);
+	resp->state->check_cancel_fn = NULL;
+	resp->check_cancel = NULL;
+
+	for (;;)
```
theory, line 210: Any chance of never reaching an exit state and getting suck in an infinite loop?

src/binary/select.c
```diff
@@ -0,0 +1,408 @@
...
+
+			/*
+			 * Quote & escape the value the way clickhouse-cpp's
+			 * WriteQuotedString did: wrap in single quotes, replace inner
+			 * specials with backslash-escapes the server's
+			 * Field::restoreFromDump understands. Without escaping inner
+			 * quotes the server stops parsing at the first `'` inside the
+			 * value, which breaks Array(String) parameters whose CH literal
+			 * already contains quoted elements.
+			 */
```
theory, line 309: Shouldn't the params implementation handle any necessary quoting? Also what if the param isn't a string? I see that the docs for `chc_query_param` say "this library passes the value through verbatim so callers keep full control of typed and non-string values", but ISTM that most database drivers I've used have data type-aware parameters. Also it's odd that the binary protocol requires quoting at all, as opposed to, I dunno, taking the binary representation of a parameter value.

src/binary/select.c
```diff
@@ -0,0 +1,408 @@
...
+ch_binary_response_free(ch_binary_response_t * resp)
+{
+	if (!resp)
+		return;
+
+	/* Avoid raising from a MemoryContextResetCallback. */
+	PG_TRY();
+	{
+		drain_until_eos(resp);
+	}
```
theory, line 411: The docs for `drain_until_eos()` say "Never raises." Is the error handling here necessary?

src/include/binary.h
```diff
@@ -0,0 +1,236 @@
...
+ * (between ch_binary_array_begin/_end) returned kind is element kind,
+ * not CHC_ARRAY.
+ */
+extern chc_kind ch_binary_column_kind(const ch_binary_insert_handle * h,
+									  size_t col);
+extern uint32_t ch_binary_column_datetime64_precision(const ch_binary_insert_handle * h,
+													  size_t col);
+
+/* Send buffered rows and clear; ready for next batch. */
+extern void ch_binary_flush_block(ch_binary_insert_handle * h);
```
theory: Aren't these internal-only?

src/include/binary.h
```diff
@@ -0,0 +1,236 @@
...
+ */
+typedef struct
+{
+	Datum	   *datums;
+	bool	   *nulls;
+	size_t		len;
+	int			ndim;			/* nesting depth, >=1 */
+	Oid			item_type;		/* leaf scalar PG type */
+	Oid			array_type;		/* PG array type (same at every level) */
+}			ch_binary_array_t;
```
theory: Private?

src/include/binary.h
```diff
@@ -0,0 +1,236 @@
...
+
+/* INSERT helpers (encode.c). */
+extern void ch_binary_prepare_insert(void *conn, const ch_query * query,
+									 ch_binary_insert_state * state);
+extern void ch_binary_insert_columns(ch_binary_insert_state * state);
+extern void ch_binary_column_append_data(ch_binary_insert_state * state, size_t colidx);
+extern void *ch_binary_make_tuple_map(TupleDesc indesc, TupleDesc outdesc, Oid relid);
+extern void ch_binary_insert_state_free(void *c);
+extern void ch_binary_do_output_conversion(ch_binary_insert_state * state,
+										   TupleTableSlot * slot);
```
theory, line 114: Private?

src/binary/convert.c
```diff
@@ -454,6 +505,16 @@ ch_binary_init_convert_state(Datum val, Oid intype, Oid outtype)
...
+		else if (intype == JSONBOID && outtype == JSONOID)
+		{
+			state->func = convert_in_jsonb_to_json;
+			state->ctype = COERCION_PATH_FUNC;
+		}
+		else if (intype == JSONOID && outtype == JSONBOID)
+		{
+			state->func = convert_in_json_to_jsonb;
+			state->ctype = COERCION_PATH_FUNC;
+		}
```
theory: Are these necessary? I thought I implemented the JSON and JSONB conversion separately so there was no need to cast (and thus copy all that data around).

src/binary/convert.c
```diff
@@ -514,12 +575,23 @@ init_output_convert_state(ch_convert_output_state * state)
...
+		state->func = convert_json_to_jsonb;
+		state->ctype = COERCION_PATH_FUNC;
+		return;
+	}
+	if (state->intype == JSONBOID && state->outtype == JSONOID)
+	{
+		state->func = convert_jsonb_to_json;
+		state->ctype = COERCION_PATH_FUNC;
 		return;
+	}
```
theory: Same question, I was sure I avoided the need for casting at all.

src/pglink.c
```diff
@@ -933,7 +927,36 @@ binary_fetch_row(ChFdwScanRowContext * ctx)
...
+		foreach(lc, attrs)
+		{
+			int			i = lfirst_int(lc);
+
+			if (state->coltypes[j] == JSONBOID &&
+				TupleDescAttr(tupdesc, i - 1)->atttypid == JSONOID)
+				state->coltypes[j] = JSONOID;
+			j++;
+		}
+	}
```
theory: Previously this happened before the call to fetch any rows. Does it now run for every row?

theory (changes requested):
SO close! Happy to help resolve some of these. Would also like to squash it into a single commit with a nice description.

src/binary/decode.c
theory: I guess we didn't have it before.

src/binary/insert.c
theory: I think, having read further, that it means the internal type for a nullable type, yes? "stripped" seems an odd turn of phrase for that.

src/binary/decode.c
theory: Yes, block by block, mentioned elsewhere.

src/binary/decode.c
serprex: not sure how we want to make the switch, but goal here is to try change as little as possible, I'm even a bit dissatisfied with error messages changing as much as they do

src/binary/encode.c
serprex: wasn't sure how best to dedupe, feels like deduping costs more lines than it saves

(tried it out, saved a couple lines)

src/binary/decode.c
serprex: checks sign bit, if negative we negate value & keep a bool to prepend `-`

C++ version was leveraging int128, while this version is having to manage individual bytes of these integers

src/binary/insert.c
serprex: insert column, will update comment

src/binary/insert.c
serprex: will consolidate on "unwrapped" so at least there's just one awkward lingo to pick up

src/binary/decode.c
serprex: yes, in future it'd make sense to generalize

src/binary/insert.c
serprex: will add `cannot insert`

src/binary/binary.c
serprex: No, `CHC_IMPLEMENTATION` goes in one file (TU) which then `#ifdef` in header includes actual function definitions

guessing VSCode is using clangd under the hood, which relies on [`compile_commands.json`](https://clang.llvm.org/docs/JSONCompilationDatabase.html), in that case use [`bear`](https://github.com/rizsotto/bear) or [`compiledb`](https://github.com/nickdiego/compiledb)

(I personally use bear, https://archlinux.org/packages/extra/x86_64/bear)

src/binary/insert.c
serprex: https://github.com/ClickHouse/pg_clickhouse/pull/254#discussion_r3293383699

`StringInfoData` also fails on 1GB+ allocations

src/binary/select.c
serprex: with a malicious ClickHouse server yes

src/binary/insert.c
serprex: We don't support inserting into `Array(Nullable(T))` currently. Will convert to error & make followup PR

src/include/binary.h
serprex: `ch_binary_make_tuple_map` is used outside binary/

src/include/binary.h
serprex: used by `src/convert.c`

src/binary/insert.c
serprex: `Array(JSON)` unsupported, same for `Array(Tuple/Map/Nested/LowCardinality/Variant/Dynamic/Point/Ring/Polygon/MultiPolygon/AggregateFunction)`

src/binary/select.c
serprex: Lies, `drain_until_eos` allocates so unless we switch to using `MCXT_ALLOC_NO_OOM` in allocator that's a problem. Updated comment

src/binary/select.c
serprex: if only ClickHouse weren't odd:
1. reads settings map:
https://github.com/ClickHouse/ClickHouse/blob/7f70b04fcdef4915ec02354449bf9b18a22220ee/src/Server/TCPHandler.cpp#L2314-L2316 *(see https://github.com/ClickHouse/ClickHouse/blob/8531400c7ee55e1002395b5f877d154122e222aa/src/Core/SettingsWriteFormat.h#L5 for settings formats, notice binary is some old deprecated thing)*
2. BaseSettings::read dispatches CUSTOM-flagged entries to SettingFieldCustom::parseFromString:
https://github.com/ClickHouse/ClickHouse/blob/7f70b04fcdef4915ec02354449bf9b18a22220ee/src/Core/BaseSettings.h#L642-L660
3. SettingFieldCustom::parseFromString calls Field::restoreFromDump:
https://github.com/ClickHouse/ClickHouse/blob/7f70b04fcdef4915ec02354449bf9b18a22220ee/src/Core/SettingsFields.cpp#L577-L580
4. Field::restoreFromDump accepts:
  - Quoted String:
https://github.com/ClickHouse/ClickHouse/blob/7f70b04fcdef4915ec02354449bf9b18a22220ee/src/Core/Field.cpp#L641-L647
  - prefixed with UInt64_, Int64_, Float64_, etc:
https://github.com/ClickHouse/ClickHouse/blob/7f70b04fcdef4915ec02354449bf9b18a22220ee/src/Core/Field.cpp#L542-L749

arguably quoting should belong in clickhouse-c given this situation, tho consumers that don't have arbitrary parameters are able to just pass constant strings. this sort of gets into the whole issue that clickhouse-c wants to just ferry bytes back & forth, with minimal encode/decode

src/binary/convert.c
serprex: removed, instead `TupleDescInitEntry` needs to just get right oid in `ch_binary_prepare_insert`

src/binary/convert.c
serprex: https://github.com/ClickHouse/pg_clickhouse/pull/254#discussion_r3314570376

src/pglink.c
serprex: yes, lifted to `binary_simple_query`

src/binary/insert.c
serprex: added comments, not sure what size of steps you're thinking

src/binary/encode.c
```diff
@@ -0,0 +1,422 @@
...
+static Oid
+ch_kind_to_pg_oid_for_insert(const chc_type * type, const char *colname)
+{
+	switch (chc_type_kind(type))
+	{
+		case CHC_INT8:
+		case CHC_INT16:
+		case CHC_UINT8:
+		case CHC_BOOL:
+			return INT2OID;
```
serprex: do we want to map `CHC_BOOL` to `BOOLOID`?

src/binary/select.c
serprex: #256 sketched read timeouts

src/binary/decode.c
theory: Yeah, I just forgot we didn't support Time already.

src/binary/decode.c
theory: Thanks for the comments. I could use a primer on how ClickHouse Decimal is represented, though not at all urgent since it works.

src/binary/decode.c
theory: `pglink.c` documented it as INT4, so this is a worthwhile fix IMO.

src/binary/encode.c
theory: It's still in `binary.c` tho?

src/binary/binary.c
theory: Seems better today; I ran `make compile_commands.json` at some point yesterday.

src/binary/insert.c
theory: Def think these Decimal conversions would be useful util functions in clickhouse-c.

src/binary/select.c
theory: Agree on that, but again, some utilities for common patterns would be super useful IMHO.

theory: Also, yeah, weird.

test/expected/import_schema.out
```diff
@@ -714,10 +715,10 @@ SELECT c1, c2, c3, c4, c5, c6, c7, NULL, c9, c10
 FROM clickhouse_bin.ints
 WHERE c1 = 127
 ORDER BY c1;
-  c1  |   c2   |     c3      |          c4          | c5  | c6 |     c7     | c8 |      c9       |           c10           
-------+--------+-------------+----------------------+-----+----+------------+----+---------------+-------------------------
- -128 | -32768 | -2147483648 | -9223372036854775808 |   0 |  0 |          0 |  0 | 1.1754942e-38 | 2.2250738585072014e-308
-  127 |  32767 |  2147483647 |  9223372036854775807 | 255 | -1 | 4294967295 |    | 3.4028233e+38 | 1.7976931348623157e+308
```
theory, line 720: Oh damn, was this the bug in the handling of int32s? Looks like it wads `65535` in the http output. If so, we should note the fix in the changelog.

theory (approved):
This is great.

test/expected/import_schema.out
theory: Added in 69c6a7e.

test/expected/binary_queries_8.out
theory, line 1: What changed to require this new file?

src/binary/encode.c
theory: We will, yes. Good exercise for me to add in a subsequent PR.

src/binary/insert.c
theory: Can add later if needed.

src/binary/encode.c
serprex: #258

test/expected/binary_queries_8.out
serprex: reran test suite, removed stale tests, this file is gone

src/binary/encode.c
serprex: defined in binary.c, used in decode/encode

theory: Oh, I thought the whole thing would be defined in the header file. Guess you can't do that with `extern`?

serprex: nope, `extern` in header is for saying some translation unit will define. you could have it only in header as `static` but then it'd be copied in every unit that includes it (& maybe optimized out of most)

theory: Yeah, no bueno

PR #257, opened 2026-05-28
Update clickhouse-cpp for proper Bool support

theory:
The new version adds a `Bool` column type and the `-DCH_MAP_BOOL_TO_UINT8=OFF` compile directive to prevent converting bools to `uint8` values. This eliminates the need for the binary driver to cast to and from `uint8` (`INT2`) to handle boolean values: we can just directly use the boolean type support.


serprex:
#254

PR #266, opened 2026-06-03
HTTP streaming: gate write_callback pause on having a row-aligned batch

JoshDreamland:
**Targets David's #220 (`streaming-fail`).**

## Root cause

`write_callback` pauses receipt whenever `write_pos >= fetch_size`, regardless of whether a complete row has landed. With small `fetch_size` (David's test uses `1`), this fires before the first `\n` arrives. The downstream chain then collapses:

- `find_batch_end` returns 0 (no row-aligned cut available).
- `pump` returns with `batch_end=0`.
- `ch_http_stream_available` returns 0.
- `ch_http_read_state_init(..., 0)` sets `state->data = NULL`.
- `http_fetch_row_from_state` returns `NULL` — which is also how it signals end-of-stream.

PG sees `NULL`, closes the cursor, and the rest of curl's pending response is discarded. Row counts varied across CI runs (3329, 3329, 3332, 3376 locally) because the loss point depends on where curl's chunk boundaries fell.

## Fix

Gate the pause on the buffer actually containing a `\n`. Same `CURL_WRITEFUNC_PAUSE` idiom; minimum-surface change (5 lines).

```c
if (self->write_pos >= (size_t) self->fetch_size
    && memchr(self->buf, '\n', self->write_pos) != NULL)
{
    self->paused = true;
    return CURL_WRITEFUNC_PAUSE;
}
```

## Test

Also extends `stream_out` with a second `SELECT *` at `fetch_size=3` so we cover a non-degenerate small-fetch case. Both pass: 3500/3500 rows.

## Alternative

Sibling PR `stream-fix-accept-then-pause` restructures `write_callback` to always accept the chunk and pause via `curl_easy_pause` afterward. Slightly larger diff, no curl-internal redelivery cycle, otherwise functionally identical. Pick one.

### DIFF
--- a/test/expected/stream_out.out
+++ b/test/expected/stream_out.out
@@ -1,5 +1,5 @@
 CREATE SERVER try_http FOREIGN DATA WRAPPER clickhouse_fdw
-    OPTIONS(dbname 'try_test', driver 'http', fetch_size '0');
+    OPTIONS(dbname 'try_test', driver 'http', fetch_size '1');
 CREATE USER MAPPING FOR CURRENT_USER SERVER try_http;
 SELECT clickhouse_raw_query('DROP DATABASE IF EXISTS try_test');
  clickhouse_raw_query 
--- a/test/expected/stream_out.out
+++ b/test/expected/stream_out.out
@@ -3534,6 +3534,13 @@ SELECT * FROM try_http_ft;
  3499
 (3500 rows)
 
+ALTER SERVER try_http OPTIONS (SET fetch_size '3');
+SELECT count(*), min(c1), max(c1), sum(c1::bigint) FROM try_http_ft;
+ count | min | max  |   sum   
+-------+-----+------+---------
+  3500 |   0 | 3499 | 6123250
+(1 row)
+
 DROP USER MAPPING FOR CURRENT_USER SERVER try_http;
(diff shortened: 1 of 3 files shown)


test/expected/stream_out.out
```diff
@@ -3534,6 +3534,13 @@ SELECT * FROM try_http_ft;
  3499
 (3500 rows)
 
+ALTER SERVER try_http OPTIONS (SET fetch_size '3');
+SELECT count(*), min(c1), max(c1), sum(c1::bigint) FROM try_http_ft;
+ count | min | max  |   sum   
+-------+-----+------+---------
+  3500 |   0 | 3499 | 6123250
+(1 row)
```
theory, line 3542: What's this for?

JoshDreamland: To add a second test that is identical to the first, except using a `fetch_size` of 3 (so that a byte or two will be read in after the newline) instead of using a size that will always lie very evenly on the boundary of one. My worry was that we just generally couldn't handle a block of data that was squarely inside a row, and that my fix ONLY worked for the first case because it landed evenly on the row boundary every time. That was not the case.

theory: Makes sense! Would you add a comment to that effect?

PR #267, opened 2026-06-03
HTTP streaming: pause receipt after the chunk that completes a batch

JoshDreamland:
**Targets David's #220 (`streaming-fail`).**

## Root cause

Same as the sibling PR. `write_callback` pauses receipt whenever `write_pos >= fetch_size`, regardless of whether a complete row has landed. With small `fetch_size` (David's test uses `1`), this fires before the first `\n` arrives. The downstream chain then collapses:

- `find_batch_end` returns 0 (no row-aligned cut available).
- `pump` returns with `batch_end=0`.
- `ch_http_stream_available` returns 0.
- `ch_http_read_state_init(..., 0)` sets `state->data = NULL`.
- `http_fetch_row_from_state` returns `NULL` — which is also how it signals end-of-stream.

PG sees `NULL`, closes the cursor, and the rest of curl's pending response is discarded.

## Fix

Restructure `write_callback` so the function does one thing per call: always append the chunk, then ask CURL to pause receipt via `curl_easy_pause(CURLPAUSE_RECV)` once the buffer contains both `>= fetch_size` bytes AND at least one `\n`. CURL delivers the *next* chunk on resume rather than redelivering the one we already hold.

```c
memcpy(self->buf + self->write_pos, contents, realsize);
self->write_pos += realsize;
self->buf[self->write_pos] = '\0';

if (self->write_pos >= (size_t) self->fetch_size
    && memchr(self->buf, '\n', self->write_pos) != NULL)
{
    self->paused = true;
    curl_easy_pause(self->curl, CURLPAUSE_RECV);
}

return realsize;
```

## Test

Also extends `stream_out` with a second `SELECT *` at `fetch_size=3` so we cover a non-degenerate small-fetch case. Both pass: 3500/3500 rows.

## Alternative

Sibling PR `stream-fix-refuse-chunk` keeps the existing `CURL_WRITEFUNC_PAUSE` idiom and just gates the refusal on the presence of a `\n` in the buffer. Smaller diff (5 lines); functionally identical. Pick one.


serprex (approved):
prefer this over #266

JoshDreamland:
This one was my preference as well. Wanted to calibrate that against  the team's.

src/http_streaming.c
```diff
@@ -217,6 +205,20 @@ write_callback(void *contents, size_t size, size_t nmemb, void *userp)
...
 	self->buf[self->write_pos] = '\0';
 
+	/*
+	 * Once we have buffered at least fetch_size bytes AND at least one
+	 * newline (so a row-aligned batch is ready), pause receipt. Pausing on
+	 * byte-count alone can starve the parser of the newline it needs when
+	 * fetch_size is small. We accept this chunk first and pause afterward
+	 * via curl_easy_pause so CURL does not redeliver bytes we already hold.
+	 */
+	if (self->write_pos >= (size_t) self->fetch_size
```
serprex, line 215: the weird case to consider is when this pauses immediately before caller would otherwise close connection. shouldn't really matter. ie, small overhead of curl pause, meaningless. assuming curl is sane when closing a paused connection

theory (approved):
Looks good. Squash it down to a single signed commit and merge.

PR #268, opened 2026-06-08
add compression option for binary protocol

serprex:
default to lz4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ All notable changes to this project will be documented in this file. It uses the
   [Semantic Versioning]: https://semver.org/spec/v2.0.0.html
     "Semantic Versioning 2.0.0"
 
+## [Unreleased]
+
+### ⚡ Improvements
+
+*   [#268] added a `compression` server option for the binary driver to enable
+    ClickHouse native protocol compression of query results and `INSERT` data.
+    Accepts `none`, `lz4`, or `zstd`, and defaults to `lz4`.
+
+  [#268]: https://github.com/ClickHouse/pg_clickhouse/pull/268"
+    "ClickHouse/pg_clickhouse#232 add compression option for binary protocol"
+
 ## [v0.3.1] — 2026-05-02
 
 This release makes binary-only changes. Once installed, any existing use of
--- a/src/binary/connection.c
+++ b/src/binary/connection.c
@@ -40,6 +40,27 @@ cancel_adapter(void *ud)
 }
 
 /*
+ * Map "compression" option to send method. NULL/unset defaults to LZ4
+ * (ClickHouse's native network default). Decode side picks codec from each
+ * frame's method byte, so callers fill both codec families regardless.
+ */
+static chc_compression
+parse_compression(const char *s)
+{
+	if (s == NULL || s[0] == '\0' || pg_strcasecmp(s, "lz4") == 0)
+		return CHC_COMP_LZ4;
+	if (pg_strcasecmp(s, "none") == 0)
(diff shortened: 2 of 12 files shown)


src/binary/connection.c
```diff
@@ -39,6 +39,27 @@ cancel_adapter(void *ud)
...
+{
+	if (s == NULL || s[0] == '\0' || pg_strcasecmp(s, "lz4") == 0)
+		return CHC_COMP_LZ4;
+	if (pg_strcasecmp(s, "none") == 0)
+		return CHC_COMP_NONE;
+	if (pg_strcasecmp(s, "zstd") == 0)
+		return CHC_COMP_ZSTD;
+
+	ereport(ERROR,
+			(errcode(ERRCODE_FDW_INVALID_OPTION_NAME),
```
theory, line 58: It annoys me that there's no `ERRCODE_FDW_INVALID_OPTION_VALUE`.

CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ All notable changes to this project will be documented in this file. It uses the
...
     "Semantic Versioning 2.0.0"
 
+## [Unreleased]
+
+### ⚡ Improvements
+
+*   Added a `compression` server option for the binary driver to enable
+    ClickHouse native protocol compression of query results and `INSERT` data.
+    Accepts `none`, `lz4`, or `zstd`, and defaults to `lz4`. Thanks to serprex
+    for the PR ([#268]).
```
theory: No more need for "thanks" now that your you're co-author. :-) Although it's handy to link to the PR.

src/binary/connection.c
```diff
@@ -236,10 +259,19 @@ ch_binary_connect(ch_connection_details * details)
 							  cancel_adapter, s);
 		}
 
+		if (comp != CHC_COMP_NONE)
+		{
+			/* both families: server may answer in a different method */
```
theory, line 264: This is confusing. Shouldn't the option be a boolean, then?

Oh, I see, we pass the compression below, eh? Could use some clarification here.

test/expected/binary_compression.out
```diff
@@ -0,0 +1,127 @@
...
+  0 | row 0
+  1 | row 1
+  2 | row 2
+  3 | row 3
+  4 | row 4
+(5 rows)
+
+-- Explicit lz4.
+CREATE SERVER comp_lz4 FOREIGN DATA WRAPPER clickhouse_fdw
+    OPTIONS(dbname 'compression_test', driver 'binary', compression 'lz4');
```
theory: I suspect you can use `ALTER SERVER` and just use the same server, no?

test/expected/binary_compression.out
```diff
@@ -0,0 +1,127 @@
...
+INSERT INTO ft_zstd VALUES (100, 'inserted');
+SELECT * FROM ft_zstd WHERE c1 = 100;
+ c1  |    c2    
+-----+----------
+ 100 | inserted
+(1 row)
+
+-- Unknown value errors when the connection is opened.
+CREATE SERVER comp_bad FOREIGN DATA WRAPPER clickhouse_fdw
+    OPTIONS(dbname 'compression_test', driver 'binary', compression 'bogus');
```
theory: I wonder if we should teach `clickhouse_fdw_validator` to inspect values, too, so it errors earlier. (No need to do here, just had to think through why it wasn't failing earlier.)

theory (approved):
🤐

src/binary/connection.c
serprex: we send lz4 or zstd, but accept either from server

PR #269, opened 2026-06-08
Add support for PostgreSQL 19beta1

theory:
Conditionally include the new `varatt.h` header so that references that `ip_family()` and friends make to `VARDATA_ANY` don't fail. This is due to some work to reduce cross-header dependencies, resulting in the need to include those that were previously implied.

Also include `access/tupconvert.h`, the new home for `execute_attr_map_tuple()`.

Move the `elmlen, bool elmbyval, char elmalign` args from `array_iter_next()` to `array_iter_setup` on Postgres 19. This changes is the result of an optimization in Postgres commit [da7a1dc0d] that allows setup to happen just once when iterating over an array. Leave the old pattern for older versions, of course.

Replace use of the `bits16` typedef, which has been removed in Postgres commit [bab2f27ea], with `uint16`, the typedef it aliased.

Add `void` as the argument list to `chfdw_get_session_settings()` and `chfdw_pushdown_regex_ok()` to silence a warning turned error by [171e67].

Add an alternate `DistinctExpr` tests to the `http.sql` and `binary_queries.sql` tests, as the original

`(c1 IS NOT NULL) IS DISTINCT FROM (c1 IS NOT NULL)` expression is now optimized away by Postgres commit [e2debb643]. Leave it for evaluation on the older versions, but also add a new quey that properly tests `IS DISTINCT FROM` pushdown.

Add new expected outputs for tests that now emit `Replaces` nodes in `EXPLAIN` output, added in Postgres commit [f2bae51df], as well as renamed `InitPlan` and `SubPlan` nodes.

New expected tests outputs for Postgres 19 have the default names; other expected outputs have had their `_\d` suffix incremented to the next versions.

Update Postgres 19 to be the standard version for all builds and tests, with only the Docker image omitted, as the base image doesn't exist yet.

  [da7a1dc0d]: https://github.com/postgres/postgres/commit/da7a1dc0d
  [bab2f27ea]: https://github.com/postgres/postgres/commit/bab2f27ea
  [171e67]:  https://github.com/ClickHouse/pg_clickhouse/commit/171e67
  [e2debb643]: https://github.com/postgres/postgres/commit/e2debb643
  [f2bae51df]: https://github.com/postgres/postgres/commit/f2bae51df

### DIFF
--- a/.github/workflows/postgres.yml
+++ b/.github/workflows/postgres.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        pg: [18, 17, 16, 15, 14, 13]
+        pg: [19, 18, 17, 16, 15, 14, 13]
         os:
           - { icon: 🐧, arch: amd64, on: ubuntu-latest }
           # - { icon: 🐧, arch: arm64, on: ubuntu-24.04-arm } # XXX
--- a/src/binary/convert.c
+++ b/src/binary/convert.c
@@ -4,6 +4,7 @@
 
 #include "funcapi.h"
 #include "access/tupdesc.h"
+#include "access/tupconvert.h"
 #include "catalog/pg_type_d.h"
 #include "catalog/pg_type.h"
 #include "utils/typcache.h"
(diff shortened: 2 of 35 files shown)


.github/workflows/postgres.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        pg: [18, 17, 16, 15, 14, 13]
+        pg: [19, 18, 17, 16, 15, 14, 13]
```
serprex, line 16: would now be a good time to drop pg13? some EOL policy would be good

src/binary/convert.c
```diff
@@ -699,26 +700,33 @@ ch_binary_do_output_conversion(ch_binary_insert_state * insert_state,
...
 					for (size_t j = 0; j < total; j++)
-					{
 						arr->datums[j] = array_iter_next(&iter, &arr->nulls[j], j,
 														 cstate->typlen, cstate->typbyval, cstate->typalign);
-					}
+#else
+					array_iter_setup(&iter, v, cstate->typlen, cstate->typbyval, cstate->typalign);
+					for (size_t j = 0; j < total; j++)
+						arr->datums[j] = array_iter_next(&iter, &arr->nulls[j], j);
+#endif
```
serprex, line 712: at citus we had a compat.h header that we'd use to define macros which would ignore removed parameters, could have a shim here that takes typlen/typbyval/typalign on both setup & next, then passes down where appropriate based on version

.github/workflows/postgres.yml
theory: Meh. Separate PR anyway.

src/binary/convert.c
theory: I thought you hated macros! 😹

serprex (approved):

PR #272, opened 2026-06-12
feat: add three-state secure option (on/off/auto) for TLS control

serprex:
rebased & extended #227


serprex:
@x4m I don't have write access to repo from #227 so made this PR instead

JoshDreamland (approved):

theory (commented):
Hrm, no test for binary servers. We should consider adding an optional test that uses ClickHouse cloud or a local TLS-enabled service like pg_stat_ch uses for testing.

PR #277, opened 2026-06-13
refactor: consolidate lookups

serprex:


src/binary/decode.c
```diff
@@ -139,55 +85,20 @@ slice_str(const chc_column * col, uint64_t row,
...
-		case CHC_DATE32:
-			return DATEOID;
-		case CHC_DATETIME:
-		case CHC_DATETIME64:
-			return TIMESTAMPTZOID;
-		case CHC_UUID:
-			return UUIDOID;
-		case CHC_IPV4:
-		case CHC_IPV6:
-			return INETOID;
```
theory, line 190: What replaces these?

theory (approved):
Much cleaner, thank you.

src/binary/decode.c
serprex: `ch_scalar_oids` array lookup, used in 2 places

PR #278, opened 2026-06-14
generate typedefs.list with find_typedef from postgres

serprex:


theory (changes requested):
I'm getting an error when I try to run it on macOS Tahoe:

```
make: *** [dev/typedefs.list] Error 1
make: *** Deleting file `dev/typedefs.list'
```

I poked at the Postgres source and it appears it keeps a [manually-maintained list of typedefs](https://github.com/postgres/postgres/blob/master/src/tools/pgindent/typedefs.list), though [020794ee4](https://github.com/postgres/postgres/commit/020794ee4) hints that there may be some automation for it. Maybe we should must maintain a manual list and periodically run `make dev/typedefs.list` on Linux to add new stuff?

Or maybe we should dump pg_bsd_indent in favor of some other linter? I like using pg_bsd_indent to keep things like Postgres itself, but maybe it doesn't matter much?

theory:
I get this error in the pgxn-tools docker image:

```
root@3e7ebfdd94d6:/repo# make dev/typedefs.list
make: *** [Makefile:125: dev/typedefs.list] Error 1
```

I get this whether or not I've installed Postgres.

serprex:
I'd vote for clang-format over pg_bsd_indent

PR #285, opened 2026-06-16
Don't use `IN` for empty array (`ANY('{}')`)

### LINKED ISSUE #275: Deparser emits empty IN () for = ANY / <> ALL over empty array constants — rejected by ClickHouse < 25.x
### Summary

When a `ScalarArrayOpExpr` with a constant empty array is pushed down, the deparser emits an empty tuple after `IN`/`NOT IN`, producing remote SQL like `(col IN ())`. Older ClickHouse servers reject this: ≤22.x with `DB::Exception: Syntax error: failed at position N (')'): ... Expected ...` (SYNTAX_ERROR), 23.x–24.x with `Function tuple requires at least one argument` (NUMBER_OF_ARGUMENTS_DOESNT_MATCH), 24.8 with `Function 'in' is supported only if second argument is constant or table expression` (UNSUPPORTED_METHOD). ClickHouse ≥ ~25.x accepts `IN ()` with correct semantics, so the failure is version-dependent on the remote side.

All references below are pinned to commit `94256f0`.

### Trigger conditions

- `SELECT * FROM ft WHERE col = ANY('{}'::int[]);` → remote `(col IN ())`
- `SELECT * FROM ft WHERE col <> ALL('{}'::int[]);` → remote `(col NOT IN ())`
- Parameterized queries `WHERE col = ANY($1)` with an empty-array bind, whenever PostgreSQL uses a **custom plan** (default for the first five executions of a prepared statement, and later whenever estimated cheaper): `eval_const_expressions` folds the extern Param into a Const, which then takes the IN path. Under a generic plan the same query takes the `has({pN:Array(T)}, col)` path with a server-side bound parameter and works — so the same application query fails intermittently depending on plan choice.

### Code walkthrough

1. `deparseScalarArrayOpExpr` (`src/deparse.c:3749`) dispatches `= ANY(Const)` (line 3772) and `<> ALL(Const)` (line 3798) to `deparseAsIn`.
2. `deparseAsIn` (`src/deparse.c:3726-3742`) appends `" IN "`/`" NOT IN "`, sets `context->array_as_tuple = true`, and deparses the array Const.
3. `deparseConst` (`src/deparse.c:2296-2298`) routes array output to `deparseArray` (`src/deparse.c:2157`); the 1-D branch (lines 2196-2218) writes `(`, loops over `nitems` elements, writes `)` — for a zero-element array this is exactly `()`. There is no empty-array guard here, unlike `deparseArrayExpr` (`src/deparse.c:3922-3932`), which CAST-wraps empty `ARRAY[]` expressions.
4. Shippability never blocks it: `T_Const` is unconditionally shippable (`src/deparse.c:362`) and the `T_ScalarArrayOpExpr` check (`src/deparse.c:484-498`) only validates the operator.

Note: a `Param` right-hand side does not hit this path (`deparseAsIn` requires `IsA(arg2, Const)`); params are sent as native ClickHouse `{pN:Type}` placeholders with `param_pN` form parts (`src/deparse.c:2431-2438`, `src/http_streaming.c:167-173`), so only plan-time constants are affected.

### Observed remote error (older ClickHouse)

```
DB::Exception: Syntax error: failed at position N (')'): ... Expected one of: <list varies by version> (SYNTAX_ERROR)
```

with the position pointing at the `)` of the empty `IN ()` mid-statement (observed in production at stable mid-statement offsets, consistent with constant-folded application queries).

### Impact

Every occurrence is a hard query failure surfaced through remote-error handling — the query never executes remotely even though the PostgreSQL-side semantics are trivial (`= ANY('{}')` is constant false, `<> ALL('{}')` constant true). Applications that build `= ANY(array)` filters from possibly-empty lists (a common ORM pattern) hit this whenever the empty case is plan-time constant and the remote server is older than ~25.x. There is currently no regression test covering empty arrays in ANY/ALL pushdown.

theory:
Prior to ClickHouse 25, an `IN` expression with no values, e.g., `WHERE x = IN()`, fails with an error. An alternative is to use `has([], x)`, which does not fail on earlier versions

Revise the code that determines when to use `IN` to detect when an `ANY` expression has passed an empty array, and if so, disqualify it for use in an `IN()` expression. This lets it continue with the `has()` function, instead. Add tests to validate this behavior, and include a note to replace the new function, `is_ok_in_expr()`, with the original, simpler `IsA(expr, Const)` expression.

Resolves #275.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,10 @@ All notable changes to this project will be documented in this file. It uses the
 *   Fixed a memory leak in the http driver when not using streaming ([#281]).
 *   Fixed a memory leak when a foreign scan repeatedly re-scans, typically a
     nested-loop join with a parameterized inner foreign scan ([#282]).
+*   Change the deparsing of `ANY()` with an empty array (`WHERE x = ANY('{}')`)
+    to a `has()` function rather than an `IN()` expression. This fixes errors
+    on versions prior to ClickHouse 25, where `IN()` with no list returns an
+    error ([#285])
 
 ### 🏗️ Build Setup
 
@@ -65,6 +69,8 @@ All notable changes to this project will be documented in this file. It uses the
   [clang-format]: https://clang.llvm.org/docs/ClangFormat.html
   [#283]: https://github.com/ClickHouse/pg_clickhouse/pull/283
     "ClickHouse/pg_clickhouse#283 clang-format"
+ [#285]: https://github.com/ClickHouse/pg_clickhouse/pull/285
+    "ClickHouse/pg_clickhouse#285 Don't use `IN` for empty array (`ANY('{}')`)"
 
 ## [v0.3.1] — 2026-05-02
 
--- a/src/deparse.c
+++ b/src/deparse.c
@@ -3853,6 +3853,39 @@ deparseAsIn(ScalarArrayOpExpr* node, deparse_expr_cxt* context, int optype) {
 }
 
 /*
+ * Returns true if `node` is a constant and, if it's output is an array, the
+ * array has more than one value. Replace with `IsA(expr, Const)` when
+ * ClickHouse 24 support dropped.
+ */
+static bool
+is_ok_in_expr(Expr* expr) {
+    Const* node;
+    Oid typoutput;
+    bool typIsVarlena;
+    AnyArrayType* array;
+
+    if (!IsA(expr, Const)) {
+        return false;
+    }
+
+    node = (Const*)expr;
+    getTypeOutputInfo(node->consttype, &typoutput, &typIsVarlena);
+    if (typoutput != F_ARRAY_OUT) {
+        return true;
+    }
+
+    /*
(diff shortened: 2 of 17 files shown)


serprex:
seems like you could replace expression with `false`

theory:
> seems like you could replace expression with `false`

Yes, I tried that and it works, but I chatted with the ClickHouse developers who said that, in the absence of benchmarking, it's likely that `IN()` expressions will have been better optimized, given that `IN()` resolves to a different function than `has()`. So I elected to keep it as-is.

serprex:
my point is that when we have `x in ()` or `has([], x)`, those could be collapsed to `false`

theory:
> my point is that when we have `x in ()` or `has([], x)`, those could be collapsed to `false`

Yeah, that was my first thought, too, though I've been leaving such optimizations to the underlying planner. Can confirm it returns `false` on both tho.

theory:
And also I wanted to minimize the work we were doing, with an eye to rip it all out when we someday drop support for ClickHouse 24.

serprex (approved):

PR #289, opened 2026-06-22
Subquery (SubPlan) pushdown

JoshDreamland:
Folds correlated/uncorrelated scalar (EXPR), EXISTS, and ANY/IN subqueries, plus NOT EXISTS/NOT IN (ANTI joins), into the remote SQL. Rebased onto current main; supersedes the retired subquery-pushdown-v3 with all review feedback folded in. Draft pending one more pass.


.github/workflows/clickhouse.yml
```diff
@@ -10,7 +10,7 @@ permissions:
 
 jobs:
   build:
-    env: { pg: 19 }
+    env: { pg: 19, CH_RELEASE: "${{ matrix.ch }}" }
```
theory: What's the reason for this change?

src/deparse.c
```diff
@@ -123,6 +148,22 @@ typedef struct deparse_expr_cxt {
 #define SUBQUERY_REL_ALIAS_PREFIX "s"
 #define SUBQUERY_COL_ALIAS_PREFIX "c"
 
+/*
+ * SubPlan-interior aliases. plan_id is unique across PlannedStmt.subplans,
+ * so q{plan_id}_{varno} can never collide with the outer query's r{N}/s{N}
+ * aliases nor with a sibling SubPlan's tables. See deparseSubPlanQuery.
```
```suggestion
 * aliases, nor with a sibling SubPlan's tables. See deparseSubPlanQuery.
```

src/deparse.c
```diff
@@ -690,8 +744,12 @@ foreign_expr_walker(Node* node, foreign_glob_cxt* glob_cxt) {
         Aggref* agg = (Aggref*)node;
         ListCell* lc;
 
-        /* Not safe to pushdown when not in grouping context */
-        if (!IS_UPPER_REL(glob_cxt->foreignrel)) {
+        /*
+         * Not safe to pushdown when not in grouping context. Inside a
+         * SubPlan's Query a bare aggregate is legal (it is the
```
```suggestion
         * SubPlan's Query a bare aggregate is legal (it's the
```

src/deparse.c
```diff
@@ -865,6 +941,237 @@ foreign_expr_walker(Node* node, foreign_glob_cxt* glob_cxt) {
...
+ * Supported SubLinkTypes:
+ *   EXPR_SUBLINK    (SELECT agg(..) ..)        scalar; aggregate-only, see
+ *                                              the single-row note below
+ *   EXISTS_SUBLINK  EXISTS (SELECT ..)
+ *   ANY_SUBLINK     x IN (SELECT ..)           single-column equality only
+ *
+ * Unsupported (and why):
+ *   ALL_SUBLINK         ClickHouse lacks a direct ALL; NOT IN arrives as
+ *                       NOT(ANY) and is handled by the BoolExpr case.
+ *   ROWCOMPARE_SUBLINK  multi-column compares not deparsed.
```
theory: Could they be in the future? Add "currently" if so.

src/deparse.c
```diff
@@ -865,6 +941,237 @@ foreign_expr_walker(Node* node, foreign_glob_cxt* glob_cxt) {
...
+ *
+ * Supported SubLinkTypes:
+ *   EXPR_SUBLINK    (SELECT agg(..) ..)        scalar; aggregate-only, see
+ *                                              the single-row note below
+ *   EXISTS_SUBLINK  EXISTS (SELECT ..)
+ *   ANY_SUBLINK     x IN (SELECT ..)           single-column equality only
+ *
+ * Unsupported (and why):
+ *   ALL_SUBLINK         ClickHouse lacks a direct ALL; NOT IN arrives as
+ *                       NOT(ANY) and is handled by the BoolExpr case.
```
theory: So can we not handle it as `NOT IN(...)`?

src/deparse.c
```diff
@@ -865,6 +941,237 @@ foreign_expr_walker(Node* node, foreign_glob_cxt* glob_cxt) {
...
+ * planner has already replaced outer Vars via SS_replace_correlation_vars);
+ * each correlation arg expression is itself checked for shippability in the
+ * OUTER scope, since deparseParam will inline it with outer aliases.
+ *
+ * Single-row note (semantic wrinkle, deliberately fenced off): a scalar
+ * subquery is required to return at most one row. When it returns ZERO rows,
+ * Postgres substitutes NULL, but ClickHouse's scalar-subquery semantics
+ * differ (it can raise or yield an empty result rather than NULL), so a naive
+ * push would change query results. We sidestep this entirely by only pushing
+ * EXPR subqueries whose top level is a bare aggregate with no GROUP BY: such a
```
theory, line 984: Could this be worked around in the future to allow pushdown?

Just in general, note which bits here are possible not not yet implemented.

src/deparse.c
```diff
@@ -865,6 +941,237 @@ foreign_expr_walker(Node* node, foreign_glob_cxt* glob_cxt) {
...
+        OpExpr* op;
+
+        if (subplan->testexpr == NULL || !IsA(subplan->testexpr, OpExpr)) {
+            return false;
+        }
+        op = (OpExpr*)subplan->testexpr;
+        if (list_length(op->args) != 2) {
+            return false;
+        }
+        if (chfdw_is_equal_op(op->opno) != 1) {
```
theory, line 1163: I wonder if we'll be able to relax this at some point; ClickHouse does allow other operators in limited conditions; from [the docs](https://clickhouse.com/docs/sql-reference/statements/select/join#on-section-conditions):

> Conditions specifying join keys must:
>
> * reference both left and right tables
> * use the equality operator
>
> Other conditions may use other logical operators but they must reference either the left or the right table of a query.

src/deparse.c
```diff
@@ -865,6 +941,237 @@ foreign_expr_walker(Node* node, foreign_glob_cxt* glob_cxt) {
...
+    /*
+     * Final gate: ClickHouse supports the correlated-subquery and NOT IN shapes
+     * this pushes down from 25.8 onward (its new analyzer), and the exact SQL
+     * we emit runs correctly there (verified against 25.8/26.3/26.5). Older
+     * servers raise an error on the pushed SQL rather than degrade, so refuse
+     * the pushdown below 25.8 and let Postgres run the SubPlan locally (correct,
+     * if slower). Checked last so a plan-time connection is opened only for a
+     * subplan that is otherwise shippable. The server version is independent of
+     * which user mapping connects, so the current user's mapping suffices to
+     * read it.
```
theory: What if the current user doesn't have a mapping? This can happen when, for example, when a query executes in a function with `SECURITY DEFINER` set to a different user. We'd be better off if we could use the same user mapping as will ultimately be used to execute the query.

src/deparse.c
```diff
@@ -1234,6 +1540,44 @@ chfdw_deparse_select_stmt_for_rel(
         quals = remote_conds;
     }
 
+    /*
+     * Detect inlined SubPlans before emitting anything: their presence forces
+     * r{N} qualification throughout the statement (see has_inlined_subplan).
+     * Quals may carry RestrictInfo decoration, which expression_tree_walker
+     * does not look through on all versions, so unwrap manually. remote_conds
+     * is checked too: for upper relations it becomes the HAVING clause.
+     */
```
theory, line 1628: Cuoldn't `is_shippable_subplan` set a property in the private info for this? Seems redundant, but maybe I'm missing something.

src/deparse.c
```diff
@@ -2625,6 +3040,205 @@ printRemotePlaceholder(Oid paramtype, int32 paramtypmod, deparse_expr_cxt* conte
...
+    }
+}
+
+/*
+ * Emit a pushed-down SubPlan subquery's SELECT list (sans the SELECT keyword).
+ *
+ * This reads like deparseExplicitTargetList but is a different beast: it
+ * consumes the SubPlan's *raw Query* targetList, so it must skip resjunk
+ * entries (a planned tlist has none), and it short-circuits EXISTS to a
+ * constant, since EXISTS ignores the select list entirely.
```
theory, line 3192: Nice.

src/deparse.c
```diff
@@ -2625,6 +3040,205 @@ printRemotePlaceholder(Oid paramtype, int32 paramtypmod, deparse_expr_cxt* conte
...
+        }
+        first = false;
+        deparseExpr(tle->expr, context);
+    }
+}
+
+/*
+ * Emit a pushed-down SubPlan subquery's FROM list (sans the FROM keyword).
+ *
+ * This looks like deparseFromExpr's job but cannot use it: there is no planned
```
```suggestion
 * This looks like deparseFromExpr but we cannot use it: there is no planned
```

test/expected/subplan_pushdown.out
```diff
@@ -0,0 +1,290 @@
...
+WHERE i.item_id = s.item_id
+  AND s.amount = (SELECT max(s2.amount) FROM sales s2
+                  WHERE s2.item_id = i.item_id)
+ORDER BY i.item_id;
+                                                                                                                                             QUERY PLAN                                                                                                                                             
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+ Foreign Scan
+   Output: i.item_id, s.amount
+   Relations: (items i) INNER JOIN (sales s)
+   Remote SQL: SELECT r1.item_id, r2.amount FROM  subplan_test.items r1 ALL INNER JOIN subplan_test.sales r2 ON (((r1.item_id = r2.item_id))) WHERE (((SELECT max(q1_1.amount) FROM subplan_test.sales q1_1 WHERE ((q1_1.item_id = (r1.item_id)))) = r2.amount)) ORDER BY r1.item_id ASC NULLS LAST
```
theory, line 153: We should also execute the queries and verify the results are correct. Same for #3 above.

test/expected/subplan_pushdown.out
```diff
@@ -0,0 +1,290 @@
...
+WHERE item_id NOT IN (SELECT item_id FROM sales WHERE region = 'east')
+ORDER BY item_id;
+ item_id 
+---------
+       4
+       6
+(2 rows)
+
+-- ============================================================
+-- 6. Scalar subquery in HAVING (TPC-H Q11 shape)
```
theory, line 233: Dupe: Q11 above.

test/expected/subquery_eq.out
```diff
@@ -193,8 +193,10 @@ $$, 'dbname=sub_eq_test');
 CREATE SCHEMA sub_eq_test;
 IMPORT FOREIGN SCHEMA sub_eq_test FROM SERVER sub_eq_svr INTO sub_eq_test;
 SET SESSION search_path = sub_eq_test,public;
--- Execute query 2.
-EXPLAIN (VERBOSE, COSTS OFF)
+-- Query 2. EXPLAIN-only: this correlated subquery executes only on ClickHouse
+-- 25.8+, so we pin the deparsed plan (version-independent) rather than the
+-- execution (which is not).
```
theory: Can we restore execution? Maybe check the version and don't run it on < 25.8.

Makefile
```diff
@@ -71,8 +71,20 @@ $(EXTENSION)-$(DISTVERSION).zip:
 	git archive-all -v --prefix "$(EXTENSION)-$(DISTVERSION)/" --force-submodules $(EXTENSION)-$(DISTVERSION).zip
 
 .PHONY: test/schedule # Depends on $(TESTS), so always rebuild.
+# The subquery-pushdown tests exercise SubPlan pushdown, which is gated on
+# ClickHouse 25.8+ (older analyzers error on the correlated SQL). When CH_RELEASE
+# names an older server, drop those tests and emit a GitHub Actions warning
+# rather than carry a second set of "not pushed down" expected files. CH_RELEASE
+# is unset for local runs and the Postgres matrix (which use the latest CH), so
+# they keep the tests.
```
theory: This confuses me. CH_RELEASE is currently set in the GitHub workflow but nowhere else. Someone downloading it and running the tests won't have it set. I don't have it set! The `Makefile` has not and should not have any idea of the ClickHouse version IMO. Nor should it have any awareness of GitHub Actions.

theory (changes requested):
Hell yeah, this is looking great. I have a few nitpicks, nothing serious. Thanks for working on this!

serprex:
re2 CI needs `libicu-dev` will make PR (#312)

theory (approved):

PR #293, opened 2026-06-23
Add ClickHouse server-version detection

JoshDreamland:
Exposes the ClickHouse server version to the planner via
`chfdw_get_server_version(UserMapping*)`, with a `chfdw_version_ge()`
comparison helper. This is the reusable plumbing a follow-up will use to gate
subquery pushdown on the server version, so older ClickHouse releases degrade
gracefully instead of erroring on SQL their analyzer cannot handle.

## What's here

- A `server_version` method on the `libclickhouse_methods` vtable, backed by a
  small dependency-free `ch_server_version` struct (shared between the
  Postgres-free driver internals and the FDW layer).
- **Binary driver:** reads the version straight from the native-protocol
  handshake (`chc_client_server_info`) — free once the connection exists.
- **HTTP driver:** ClickHouse exposes no version over HTTP (no response header,
  and the "connection" is just a curl handle with no server round-trip), so it
  fetches `SELECT version()` once and caches it on the connection.
- A `clickhouse_server_version(text)` SQL function returning
  `major.minor.patch`, which exercises both drivers and is covered by the new
  `server_version` regression test.

## Notes

- No behavior change to existing flows: nothing calls the getter at plan time
  yet, so planning still opens no connection. The (amortized, once-per-session)
  plan-time connect arrives with the gate in the follow-up PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,13 @@ All notable changes to this project will be documented in this file. It uses the
   [Semantic Versioning]: https://semver.org/spec/v2.0.0.html
     "Semantic Versioning 2.0.0"
 
-## [0.3.3] — Unreleased
+## [0.4.0] — Unreleased
 
 ### ⚡ Improvements
 
+*   Added the `clickhouse_server_version(server)` function, which reports the
+    ClickHouse server version ("major.minor.patch") for a foreign server
+    ([#293]).
 *   Added pushdown for more aggregate functions ([#290]):
     *   [corr]
     *   [covarpop]
@@ -57,7 +60,7 @@ All notable changes to this project will be documented in this file. It uses the
     Postgres ordered set aggregate functions, putting them under their own
     header, "Custom Ordered Set Aggregates".
 
-  [v0.3.3]: https://github.com/ClickHouse/pg_clickhouse/compare/v0.3.2...v0.3.3
+  [v0.4.0]: https://github.com/ClickHouse/pg_clickhouse/compare/v0.3.2...v0.4.0
   [#290]: https://github.com/ClickHouse/pg_clickhouse/pull/290
     "ClickHouse/pg_clickhouse#290 Add pushdown for statistical aggregate functions"
   [corr]: https://clickhouse.com/docs/sql-reference/aggregate-functions/reference/corr
@@ -69,6 +72,8 @@ All notable changes to this project will be documented in this file. It uses the
   [var_samp/variance]: https://clickhouse.com/docs/sql-reference/aggregate-functions/reference/varSamp
   [#291]: https://github.com/ClickHouse/pg_clickhouse/pull/291
     "ClickHouse/pg_clickhouse#291 Push down ordered set aggregate functions"
+  [#293]: https://github.com/ClickHouse/pg_clickhouse/pull/293
+    "ClickHouse/pg_clickhouse#293 Add ClickHouse server-version detection plumbing"
   [#296]: https://github.com/ClickHouse/pg_clickhouse/pull/296
     "ClickHouse/pg_clickhouse#296 Fix benchmark queries that crash/hang with binary driver"
   [#300]: https://github.com/ClickHouse/pg_clickhouse/pull/300
--- a/doc/pg_clickhouse.md
+++ b/doc/pg_clickhouse.md
@@ -1047,6 +1047,27 @@ SELECT clickhouse_raw_query(
 (1 row)
 ```
 
+#### `clickhouse_server_version`
+
+```sql
+SELECT clickhouse_server_version('taxi_srv');
+```
+
+Report the ClickHouse server version, as `major.minor.patch`, for the named
+foreign server, connecting if necessary using the server's options and the
+current user's user mapping:
+
+```sql
+ clickhouse_server_version
+---------------------------
+ 25.8.1
+(1 row)
+```
+
+Reads the version from the native-protocol connection handshake, or over HTTP
+from a single `SELECT version()` query, and caches it for the life of the
+connection.
+
 ### Pushdown Functions
 
 `pg_clickhouse` pushes down a subset of the PostgreSQL builtin functions used
--- a/src/binary/connection.c
+++ b/src/binary/connection.c
@@ -358,6 +358,37 @@ ch_binary_is_broken(const ch_binary_connection_t* conn) {
     return s ? s->broken : false;
 }
 
+/*
+ * Reports the ClickHouse server version (major.minor.patch) for a binary
+ * connection. The native protocol reports the server version during the
+ * handshake, so it reads the cached handshake info rather than issuing a
+ * query. Writes 0 to all out-params when the version is unavailable.
+ */
+void
+ch_binary_server_version(
+    const ch_binary_connection_t* conn,
+    int* major,
+    int* minor,
+    int* patch
+) {
+    *major = *minor = *patch = 0;
+    if (!conn) {
+        return;
+    }
+
+    const struct ch_binary_state* s = (const struct ch_binary_state*)conn->client;
+    if (!s || !s->client) {
+        return;
+    }
+
+    const chc_server_info* info = chc_client_server_info(s->client);
+    if (info) {
+        *major = (int)info->version_major;
+        *minor = (int)info->version_minor;
+        *patch = (int)info->version_patch;
+    }
+}
+
 void
 ch_binary_close(ch_binary_connection_t* conn) {
     if (!conn) {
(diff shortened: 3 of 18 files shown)


serprex (approved):
needs CHANGELOG entry

src/http.c
```diff
@@ -217,6 +217,67 @@ ch_http_simple_query(ch_http_connection_t* conn, const ch_query* query) {
...
+    if (conn == NULL) {
+        return;
+    }
+
+    if (conn->version.major == 0) {
+        /*
+         * Build the query directly rather than via new_query(): this file does
+         * not include fdw.h (which declares chfdw_get_session_settings), and the
+         * version probe needs no session settings.
+         */
```
serprex:
```suggestion
```

theory (commented):
Will read carefully tomorrow, but just to note here, if we're adding a new SQL function, we need to increment to v0.4.0 and add an upgrade script, `sql/pg_clickhouse--0.3--0.4.sql`. See #196 for a pervious such change to the extension's catalog.

theory (changes requested):

theory:
Simpler example of incrementing the version and adding an upgrade script in ClickHouse/pg_re2#22

src/binary/connection.c
```diff
@@ -358,6 +358,31 @@ ch_binary_is_broken(const ch_binary_connection_t* conn) {
     return s ? s->broken : false;
 }
 
+void
```
theory, line 367: Please add a doc comment.

src/include/fdw.h
```diff
@@ -111,6 +115,18 @@ ch_connection
...
+ * user mapping, connecting if necessary. Safe to call during planning.
+ * Returns {0, 0, 0} when the driver cannot report a version.
+ */
+ch_server_version
+chfdw_get_server_version(UserMapping* user);
+/* True if version v is at least major.minor. */
+static inline bool
+chfdw_version_ge(ch_server_version v, int major, int minor) {
+    return v.major > major || (v.major == major && v.minor >= minor);
+}
```
theory: Maybe this should live in server_version.h?

src/http.c
```diff
@@ -217,6 +217,67 @@ ch_http_simple_query(ch_http_connection_t* conn, const ch_query* query) {
...
+        ch_query query           = { "SELECT version()", 0, NULL, NULL, NULL, NULL };
+        ch_http_response_t* resp = ch_http_simple_query(conn, &query);
+
+        if (resp != NULL) {
```
theory, line 238: We should log an error if the response is an error.

src/http.c
```diff
@@ -217,6 +217,67 @@ ch_http_simple_query(ch_http_connection_t* conn, const ch_query* query) {
...
+ * on the connection, so only the first call issues a query.
+ */
+void
+ch_http_server_version(ch_http_connection_t* conn, int* major, int* minor, int* patch) {
+    *major = *minor = *patch = 0;
+    if (conn == NULL) {
+        return;
+    }
+
+    if (conn->version.major == 0) {
```
theory, line 234: Do we know for sure that `major` was allocated with all memory set to 0?

theory (commented):
This looks good, though it doesn't look like the http connection caches the version info to allow multiple calls with the same connection without each resulting in an HTTP request.

Please add the new function to `pg_clickhouse.md` and an item to `CHANGELOG.md`, as well as the aforementioned migration script and version increments.

Also I thought ClickHouse returned a header with the version, but apparently not:

```console
❯ curl -iX POST 'http://localhost:8123/' -d 'SELECT 1'
HTTP/1.1 200 OK
X-ClickHouse-Summary: {"read_rows":"1","read_bytes":"1","written_rows":"0","written_bytes":"0","total_rows_to_read":"1","result_rows":"0","result_bytes":"0","elapsed_ns":"5185292"}
Date: Wed, 24 Jun 2026 19:58:06 GMT
Connection: Keep-Alive
Content-Type: text/tab-separated-values; charset=UTF-8
Access-Control-Expose-Headers: X-ClickHouse-Query-Id,X-ClickHouse-Summary,X-ClickHouse-Server-Display-Name,X-ClickHouse-Format,X-ClickHouse-Timezone,X-ClickHouse-Exception-Code
X-ClickHouse-Server-Display-Name: clickhouse01
Transfer-Encoding: chunked
X-ClickHouse-Query-Id: 93bf9289-6dbf-4621-a5d9-9c183e8d8484
X-ClickHouse-Format: TabSeparated
X-ClickHouse-Timezone: UTC
Keep-Alive: timeout=3, max=9999
```

test/sql/server_version.sql
```diff
@@ -0,0 +1,27 @@
...
+    OPTIONS (driver 'binary');
+CREATE USER MAPPING FOR CURRENT_USER SERVER srv_ver_binary;
+
+CREATE SERVER srv_ver_http FOREIGN DATA WRAPPER clickhouse_fdw
+    OPTIONS (driver 'http');
+CREATE USER MAPPING FOR CURRENT_USER SERVER srv_ver_http;
+
+SELECT clickhouse_server_version('srv_ver_binary') ~ '^[0-9]+\.[0-9]+\.[0-9]+$'
+           AS binary_well_formed,
+       split_part(clickhouse_server_version('srv_ver_binary'), '.', 1)::int >= 1
```
theory: This seems redundant, the regex test covers it.

src/http.c
JoshDreamland: Yes; it's `calloc`'d to start and then I specially handle that failure mode and zero out the major version if I don't read at least two of the four components.

src/binary/connection.c
JoshDreamland: Added.

src/http.c
JoshDreamland: Deleted.

src/http.c
JoshDreamland: Logged.

doc/pg_clickhouse.md
```diff
@@ -1043,6 +1043,27 @@ SELECT clickhouse_raw_query(
...
+```sql
+ clickhouse_server_version
+---------------------------
+ 25.8.1
+(1 row)
+```
+
+The version is read from the connection — over the native protocol it comes
+from the handshake, and over the HTTP interface it is obtained with a single
+`SELECT version()` and cached for the life of the connection.
```
theory: Fuck the passive voice.

```suggestion
Reads the version from the native protocol connection handshake or from a
single `SELECT version()` http query and cached for the life of the
connection.
```

src/binary/connection.c
```diff
@@ -358,6 +358,37 @@ ch_binary_is_broken(const ch_binary_connection_t* conn) {
     return s ? s->broken : false;
 }
 
+/*
+ * Report the ClickHouse server version (major.minor.patch) for a binary
+ * connection. The native protocol reports the server version during the
+ * handshake, so this reads the cached handshake info rather than issuing a
```
theory:
```suggestion
 * handshake, so read the cached handshake info rather than issuing a
```

src/http.c
```diff
@@ -217,6 +217,72 @@ ch_http_simple_query(ch_http_connection_t* conn, const ch_query* query) {
     return resp;
 }
 
+/*
+ * Fetch and cache the ClickHouse server version via SELECT version(). Writes 0
+ * to all out-params when the version cannot be determined. The result is cached
+ * on the connection, so only the first call issues a query.
```
theory:
```suggestion
 * to all out-params when the version cannot be determined. Caches the result
 * on the connection, so only the first call issues a query.
```

CHANGELOG.md
```diff
@@ -7,10 +7,13 @@ All notable changes to this project will be documented in this file. It uses the
   [Semantic Versioning]: https://semver.org/spec/v2.0.0.html
     "Semantic Versioning 2.0.0"
 
-## [0.3.3] — Unreleased
+## [0.4.0] — Unreleased
 
 ### ⚡ Improvements
 
+*   Added a `clickhouse_server_version(server)` function reporting the
```
```suggestion
*   Added the `clickhouse_server_version(server)` function, which reports the
```

JoshDreamland (commented):
Re:caching: this info lives on the connection itself (`conn->version`), yes, but specifically in the long-lived "gate" connection that all our scans use. So it's only re-fetched when a connection is re-made, which is the right behavior anyway or we could be connecting to a different version without knowing. You don't have to worry about it querying the version for each scan, though.

JoshDreamland:
Comments updated; active voice, present indicative.

theory (approved):
Brilliant!

PR #296, opened 2026-06-25
Fix benchmark queries that crash/hang with binary driver

JoshDreamland:
clickhouse-cpp read eagerly (buffered the whole result, freeing the
connection immediately), while clickhouse-c reads lazily (streams one
block at a time, holding connection open through iteration).

Two independent bugs, both in the FDW layer (binary driver untouched):

1. Concurrent foreign scans no longer share one connection.
   The binary driver streams results, holding a connection open through
   iteration, and ClickHouse's native protocol has no cursors to use to
   multiplex result sets on a single connection. So two scans live at
   once (a correlated subquery's inner + outer; a nestloop's inner)
   collided and crashed.

2. Had a `batch_cxt` use-after-free on rescan.
   ReScanForeignScan was wired to EndForeignScan, which deleted the
   per-scan `batch_cxt` and never recreated it, so the next
   `IterateForeignScan` built its cursor and response in freed memory.
   This bug exists on main and is driver-agnostic. It only hangs with
   clickhouse-c backing, whose cursor teardown deletes a nested context
   (`resp->cxt`) over the freed arena and loops; the buffering drivers
   (clickhouse-cpp before, HTTP today) don't do this during teardown.
   With clickhouse-c, the connection collision crashed first, further
   hiding this bug.

A new `ch_scan_connection` handle leases the cached connection to the
first scan and hands any scan that begins in the meantime its own
private connection; release closes private connections and only drops
the busy lease on the cached one (the connection cache owns the cached
connection's lifetime, not the scan). A transaction-end callback
reclaims private connections left by error paths.

Split `ReScanForeignScan` and `EndForeignScan`: `ReScan` resets
`batch_cxt` (keeps it live for the next iteration); `End` deletes it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,16 @@ All notable changes to this project will be documented in this file. It uses the
     *   `percentile_disc(double)` → `quantileExactLow()`
     *   `percentile_disc(double[])` → `quantilesExactLow()`
 
+### 🐞 Bug Fixes
+
+*   Fixed crashes when a single query ran more than one foreign scan on the
+    binary (native-protocol) driver at once, such as a correlated subquery or a
+    nested-loop join over foreign tables, colliding in the single connection.
+    Each concurrently active scan now gets its own connection
+    ([#296]).
+*   Fixed a use-after-free of a foreign scan's batch memory context on rescan
+    that could corrupt memory and hang ([#296]).
+
 ### 📚 Documentation
 
 *   Separated the list of custom ordered set aggregate functions provided by
@@ -44,6 +54,8 @@ All notable changes to this project will be documented in this file. It uses the
   [var_samp/variance]: https://clickhouse.com/docs/sql-reference/aggregate-functions/reference/varSamp
   [#291]: https://github.com/ClickHouse/pg_clickhouse/pull/291
     "ClickHouse/pg_clickhouse#291 Push down ordered set aggregate functions"
+  [#296]: https://github.com/ClickHouse/pg_clickhouse/pull/296
+    "ClickHouse/pg_clickhouse#296 Fix benchmark queries that crash/hang with binary driver"
 
 ## [v0.3.2] — 2026-06-16
 
--- a/src/connection.c
+++ b/src/connection.c
@@ -35,8 +35,19 @@
  * Connection cache (initialized on first use)
  */
 static HTAB* ConnectionHash = NULL;
+
+/*
+ * Private (non-cached) connections handed to foreign scans that needed a
+ * connection while the cached one was already serving a live ancestor scan.
+ * Tracked so a transaction-end callback can close any whose scan did not
+ * release cleanly (e.g. on error).
+ */
+static List* PrivateConnections = NIL;
+
 static void
 chfdw_inval_callback(Datum arg, int cacheid, uint32 hashvalue);
+static void
+chfdw_xact_callback(XactEvent event, void* arg);
 
 static ch_connection
 clickhouse_connect(ForeignServer* server, UserMapping* user) {
@@ -83,8 +94,8 @@ clickhouse_connect(ForeignServer* server, UserMapping* user) {
     }
 }
 
-ch_connection
-chfdw_get_connection(UserMapping* user) {
+static ConnCacheEntry*
+get_connection_entry(UserMapping* user) {
     bool found;
     ConnCacheEntry* entry;
     ConnCacheKey key;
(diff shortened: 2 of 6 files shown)


src/connection.c
```diff
@@ -178,7 +190,76 @@ chfdw_get_connection(UserMapping* user) {
...
+    return get_connection_entry(user)->gate;
+}
+
+ch_scan_connection
+chfdw_get_scan_connection(UserMapping* user) {
+    ConnCacheEntry* entry = get_connection_entry(user);
+    ch_scan_connection sconn;
+
+    if (!entry->busy) {
+        entry->busy      = true;
```
theory, line 207: Is there a race condition here between checking if it's busy and setting it to true?

src/fdw.c
```diff
@@ -3390,7 +3439,7 @@ clickhouse_fdw_handler(PG_FUNCTION_ARGS) {
     routine->GetForeignPlan     = clickhouseGetForeignPlan;
     routine->BeginForeignScan   = clickhouseBeginForeignScan;
     routine->IterateForeignScan = clickhouseIterateForeignScan;
-    routine->ReScanForeignScan  = clickhouseEndForeignScan;
+    routine->ReScanForeignScan  = clickhouseRedoForeignScan;
```
theory: Please call it `clickhouseReScanForeignScan` so that it's the same name as the field and aligns with the name of all the other callback hooks.

src/fdw.c
```diff
@@ -1221,6 +1229,46 @@ clickhouseIterateForeignScan(ForeignScanState* node) {
...
+     * Reset, not delete: a redo is followed by more IterateForeignScan calls
+     * that switch into batch_cxt to build the next cursor and response, so the
+     * context must stay live. The cursor and response contexts are children of
+     * batch_cxt — deleting it here would leave the next iteration allocating in
+     * a freed context, corrupting the response context and hanging its teardown
+     * on the following redo of a correlated subquery.
+     */
+    if (fsstate && fsstate->batch_cxt) {
+        MemoryContextReset(fsstate->batch_cxt);
+    }
```
theory: Why wouldn't this go into `drop_scan_cursor()`?

src/fdw.c
```diff
@@ -1230,12 +1278,13 @@ clickhouseEndForeignScan(ForeignScanState* node) {
...
     time_used = 0;
+    drop_scan_cursor(fsstate);
     if (fsstate) {
-        if (fsstate->ch_cursor) {
-            MemoryContextDelete(fsstate->ch_cursor->memcxt);
-            fsstate->ch_cursor = NULL;
+        if (fsstate->batch_cxt) {
+            MemoryContextDelete(fsstate->batch_cxt);
+            fsstate->batch_cxt = NULL;
         }
```
theory: Why wouldn't this go into `drop_scan_cursor()`? Does it really need to be set to `NULL`? At a glance, I don't see that pattern after any other calls to `MemoryContextDelete()`.

theory (changes requested):
This seems fine, modulo a few questions. More broadly, though, I wonder if we shouldn't use a connection pool of some kind, instead.

theory:
Please also add items describing the fixed bugs to `CHANGELOG.md`.

theory:
Ran `dev/tpch` and confirmed this fixes the errors. Can we add a regression test, too?

src/connection.c
JoshDreamland: No concurrency; this is for nested/interleaved scans. The existence of another consumer is a property of nested execution.

src/fdw.c
JoshDreamland: I can move it into there now, sure, but it will need a conditional. Have a look.

src/fdw.c
JoshDreamland: As above. And no, it doesn't need to be set that way, though doing so makes it obvious that it's not a valid handle as opposed to leaving a poisoned handle in an object that outlives this routine. I agree with removing it for efficiency, though.

src/fdw.c
JoshDreamland: Done.

src/connection.c
theory: Maybe add a comment to that effect.

src/fdw.c
theory: I like this better, the logic is all in one place so we're always doing it right.

CHANGELOG.md
```diff
@@ -25,6 +25,16 @@ All notable changes to this project will be documented in this file. It uses the
     *   `percentile_disc(double)` → `quantileExactLow()`
     *   `percentile_disc(double[])` → `quantilesExactLow()`
 
+### 🐞 Bug Fixes
+
+*   Fixed crashes when a single query ran more than one foreign scan on the
+    binary (native-protocol) driver at once — such as a correlated subquery or a
+    nested-loop join over foreign tables — which previously collided on one
+    shared connection. Each concurrently active scan now gets its own connection
+    ([#296]).
```
theory, line 34: In general let's edit to sound less like the artificial text extruder.

```suggestion
*   Fixed crashes when a single query ran more than one foreign scan on the
    binary (native-protocol) driver at once, such as a correlated subquery or a
    nested-loop join over foreign tables, colliding in the single connection.
    Each concurrently active scan now gets its own connection ([#296]).
```

theory (approved):

PR #297, opened 2026-06-26
Shorten timeout for binary driver in query_cancel

theory:
It was failing with `canceling statement due to statement timeout` rather than the expected `pg_clickhouse: query was canceled`.


serprex (approved):

serprex:
#304

PR #298, opened 2026-06-27
Simple partitioned aggregation

serprex:
Implements support for pushing down some partial aggregates. This opens up having partitioned tables with mix of postgres/clickhouse partitions. Useful for example where old data is dropped from postgres, but kept in ClickHouse

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,10 +43,19 @@ All notable changes to this project will be documented in this file. It uses the
     bounding memory for large `COPY FROM` and `INSERT SELECT` ([#303]).
 *   Added pushdown for the three-argument forms of `ltrim`, `rtrim`, and
     `btrim` ([#307]).
-*   Added support for binary driver to `clickhouse_raw_query()`.
+*   Added support for binary driver to `clickhouse_raw_query()` ([#309]).
 *   Added `clickhouse_query(server, sql)`, a set-returning function that runs a
     query against a configured foreign server and returns its rows typed by the
-    caller's column definition list.
+    caller's column definition list ([#309]).
+*   Added pushdown support for partial aggregates under partitionwise
+    aggregation, so a query over a partitioned table mixing local and foreign
+    partitions computes the foreign partition's aggregate on ClickHouse
+    instead of fetching its rows. Covers decomposable aggregates (`count`,
+    `sum`, `min`, `max`, `bool_and`/`bool_or`, `bit_and`/`bit_or`), plus `avg`
+    over integers and `avg`/`var_pop`/`var_samp`/`stddev_pop`/`stddev_samp`
+    over floats. Aggregates with an internal transition state (anything over
+    `numeric`, `avg(bigint)`, `avg(interval)`) still fall back to local
+    aggregation. Requires `enable_partitionwise_aggregate` ([#298]).
 
 ### 🐞 Bug Fixes
 
@@ -84,6 +93,8 @@ All notable changes to this project will be documented in this file. It uses the
     "ClickHouse/pg_clickhouse#293 Add ClickHouse server-version detection plumbing"
   [#296]: https://github.com/ClickHouse/pg_clickhouse/pull/296
     "ClickHouse/pg_clickhouse#296 Fix benchmark queries that crash/hang with binary driver"
+  [#298]: https://github.com/ClickHouse/pg_clickhouse/pull/298
+    "ClickHouse/pg_clickhouse#296 Simple partitioned aggregation"
   [#300]: https://github.com/ClickHouse/pg_clickhouse/pull/300
     "ClickHouse/pg_clickhouse#300 fix(http): handle subsecond precision"
   [#301]: https://github.com/ClickHouse/pg_clickhouse/pull/301
@@ -92,6 +103,8 @@ All notable changes to this project will be documented in this file. It uses the
     "ClickHouse/pg_clickhouse#303 Flush buffered data during binary insert"
   [#307]: https://github.com/ClickHouse/pg_clickhouse/pull/307
     "ClickHouse/pg_clickhouse#307 Push down three-argument trim functions"
+  [#307]: https://github.com/ClickHouse/pg_clickhouse/pull/309
+    "ClickHouse/pg_clickhouse#309 add binary support to clickhouse_raw_query, add clickhouse_query"
 
 ## [v0.3.2] — 2026-06-16
 
--- a/doc/pg_clickhouse.md
+++ b/doc/pg_clickhouse.md
@@ -577,6 +577,83 @@ try=# EXPLAIN (ANALYZE, VERBOSE)
  the number of rows that must be pulled back into Postgres from 1000 (all of
  them) to just 8, one for each node.
 
+### Partitioned Tables
+
+A PostgreSQL [partitioned table] can mix local partitions with foreign
+partitions backed by ClickHouse. A common layout offloads older data to
+ClickHouse while recent data stays in PostgreSQL:
+
+```pgsql
+CREATE TABLE events (id int, ts date, val int, amt float8)
+    PARTITION BY RANGE (ts);
+
+-- 2023 data lives on ClickHouse
+CREATE FOREIGN TABLE events_2023 PARTITION OF events
+    FOR VALUES FROM ('2023-01-01') TO ('2024-01-01')
+    SERVER ch_svr OPTIONS (table_name 'events');
+
+-- 2024 data stays local
+CREATE TABLE events_2024 PARTITION OF events
+    FOR VALUES FROM ('2024-01-01') TO ('2025-01-01');
+```
+
+For example on how to move data from local to foreign partitions, see
+[offload-partition.sql](offload-partition.sql).
+
+Aggregates spanning both local and foreign partitions need [partitionwise
+aggregation], which PostgreSQL disables by default:
+
+```pgsql
+SET enable_partitionwise_aggregate = on;
+```
+
+With `enable_partitionwise_aggregate` enabled, PostgreSQL computes a *partial
+aggregate* below `Append`, then a *finalize aggregate* above combines those
+partials into result. pg_clickhouse pushes the foreign partition's partial
+down to ClickHouse:
+
+```pgsql
+try=# EXPLAIN (VERBOSE, COSTS OFF)
+       SELECT count(*), sum(val), min(ts), max(ts) FROM events;
+                                                       QUERY PLAN
+-------------------------------------------------------------------------------------------------------------------------
+ Finalize Aggregate
+   Output: count(*), sum(events.val), min(events.ts), max(events.ts)
+   ->  Append
+         ->  Foreign Scan
+               Output: (PARTIAL count(*)), (PARTIAL sum(events.val)), (PARTIAL min(events.ts)), (PARTIAL max(events.ts))
+               Relations: Aggregate on (events_2023 events)
+               Remote SQL: SELECT count(*), sum(val), min(ts), max(ts) FROM "default".events
+         ->  Partial Aggregate
+               Output: PARTIAL count(*), PARTIAL sum(events_1.val), PARTIAL min(events_1.ts), PARTIAL max(events_1.ts)
+               ->  Seq Scan on public.events_2024 events_1
+                     Output: events_1.val, events_1.ts
+```
+
+#### When partial aggregates push down
+
+PostgreSQL represents a partial aggregate as a *transition state* that the
+finalize step combines across partitions. pg_clickhouse can push a partition's
+partial down only when it can express it as a ClickHouse value:
+
+*   **Decomposable aggregates** whose transition state is already the final
+    value, push down directly: `count`, `sum`, `min`, `max`, `bool_and`/`every`,
+    `bool_or`, `bit_and`, `bit_or`, and `bit_xor`.
+*   **`avg` over integers** pushes its `{count, sum}` state as an array.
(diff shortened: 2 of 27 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,15 @@ All notable changes to this project will be documented in this file. It uses the
     *   `percentile_cont(double[])` → `quantiles()`
     *   `percentile_disc(double)` → `quantileExactLow()`
     *   `percentile_disc(double[])` → `quantilesExactLow()`
+*   Push down partial aggregates under partitionwise aggregation, so a query
```
theory: CHANGELOG is usually past tense, e.g., "Added…"

src/deparse.c
```diff
@@ -382,6 +382,67 @@ chfdw_is_equal_op(Oid opno) {
...
+agg_partial_kind(Aggref* agg) {
+    HeapTuple tuple;
+    Form_pg_aggregate aggform;
+    AggPartialKind kind = AGG_PARTIAL_NONE;
+
+    if (agg->aggdistinct) {
+        return AGG_PARTIAL_NONE;
+    }
+
+    tuple = SearchSysCache1(AGGFNOID, ObjectIdGetDatum(agg->aggfnoid));
```
theory, line 416: No cache miss because we know it has already been created by the planner, yes? Maybe comment as much.

src/deparse.c
```diff
@@ -382,6 +382,67 @@ chfdw_is_equal_op(Oid opno) {
...
+    aggform = (Form_pg_aggregate)GETSTRUCT(tuple);
+
+    if (!OidIsValid(aggform->aggcombinefn)) {
+        kind = AGG_PARTIAL_NONE;
+    } else if (!OidIsValid(aggform->aggfinalfn) && !OidIsValid(aggform->aggserialfn)) {
+        kind = AGG_PARTIAL_DIRECT;
+    } else if (
+        !OidIsValid(aggform->aggserialfn) && !agg->aggvariadic &&
+        agg->aggorder == NIL && agg->aggkind == AGGKIND_NORMAL
+    ) {
```
theory, line 433: Please add comments explaining what these conditionals check in English.

src/deparse.c
```diff
@@ -4224,6 +4286,85 @@ aggref_on_aggregate_function(Aggref* node, deparse_expr_cxt* context) {
...
+
+    /* Capture FILTER condition as each component's -If argument. */
+    if (node->aggfilter) {
+        initStringInfo(&condbuf);
+        context->buf = &condbuf;
+        deparseExpr((Expr*)node->aggfilter, context);
+        cf = psprintf(", (%s) > 0", condbuf.data);
+    }
+
+    context->buf = buf;
```
theory, line 4336: I assume nothing above throws an exception so this always executes, or if it does throw an exception the buffer won't be used anyway.

src/deparse.c
```diff
@@ -4224,6 +4286,85 @@ aggref_on_aggregate_function(Aggref* node, deparse_expr_cxt* context) {
...
+            cf,
+            ifSuffix,
+            arg,
+            cf,
+            ifSuffix,
+            arg,
+            cf,
+            ifSuffix,
+            arg,
+            cf
```
theory, line 4375: Okay, the clang-format a _little_ silly sometimes 😂

src/deparse.c
```diff
@@ -4224,6 +4286,85 @@ aggref_on_aggregate_function(Aggref* node, deparse_expr_cxt* context) {
...
+        );
+    } else {
+        appendStringInfo(
+            buf,
+            "[toFloat64(count%s(%s%s)), sum%s(toFloat64(%s)%s), "
+            "if(count%s(%s%s) > 0, sum%s(pow(toFloat64(%s), 2)%s) - "
+            "pow(sum%s(toFloat64(%s)%s), 2) / count%s(%s%s), 0)]",
```
theory, line 4357: What is this about?

test/expected/partitioning_http.out
```diff
@@ -0,0 +1,170 @@
+SET datestyle = 'ISO';
+SET max_parallel_workers_per_gather = 0;
+CREATE SERVER pwagg_svr FOREIGN DATA WRAPPER clickhouse_fdw
+    OPTIONS(dbname 'pwagg_test', driver 'http');
```
theory, line 4: Should we standardize on binary?

test/expected/partitioning_http.out
```diff
@@ -0,0 +1,170 @@
...
+-- partition while the hot 2024 range stays local: the layout a consumer builds
+-- when offloading old partitions (offload itself is left to the consumer)
+CREATE TABLE events (id int, ts date, val int, amt float8) PARTITION BY RANGE (ts);
+CREATE FOREIGN TABLE events_cold PARTITION OF events
+    FOR VALUES FROM ('2023-01-01') TO ('2024-01-01')
+    SERVER pwagg_svr OPTIONS (table_name 'events');
+CREATE TABLE events_hot PARTITION OF events
+    FOR VALUES FROM ('2024-01-01') TO ('2025-01-01');
+INSERT INTO events_hot VALUES (100,'2024-01-10',5,5), (101,'2024-02-15',15,15);
+SET enable_partitionwise_aggregate = on;
```
theory, line 41: Surprised this isn't the default. I think we need to add documentation about the partitioning behavior and recommendations such as this.

test/expected/partitioning_http.out
```diff
@@ -0,0 +1,170 @@
...
+   ->  Append
+         ->  Foreign Scan
+               Output: (PARTIAL count(*)), (PARTIAL sum(events.val)), (PARTIAL min(events.ts)), (PARTIAL max(events.ts))
+               Relations: Aggregate on (events_cold events)
+               Remote SQL: SELECT count(*), sum(val), min(ts), max(ts) FROM pwagg_test.events
+         ->  Partial Aggregate
+               Output: PARTIAL count(*), PARTIAL sum(events_1.val), PARTIAL min(events_1.ts), PARTIAL max(events_1.ts)
+               ->  Seq Scan on public.events_hot events_1
+                     Output: events_1.val, events_1.ts
+(11 rows)
```
theory, line 57: Fantastic!

src/deparse.c
```diff
@@ -4224,6 +4286,85 @@ aggref_on_aggregate_function(Aggref* node, deparse_expr_cxt* context) {
     return found;
 }
 
+/*
+ * Emit partial aggregate as a ClickHouse array holding Postgres
+ * transition state, so local Finalize combines across partitions.
```
theory, line 4304: And this just works, it's what the finalize functions expect?

theory (changes requested):
This is SPLENDID, thank you! Just a few things to tweak and some questions. Documentation to explain how this works and when it doesn't work could be a different PR, as long as it happens. I want it to be clear to people why some aggregate across partitions will or won't map/reduce.

src/deparse.c
serprex: yes, this is why we're so limited. we don't support aggregates whose partial aggregation type is `internal`

src/deparse.c
serprex: [we're doing the naive schoolbook way](https://github.com/postgres/postgres/blob/f5cc81719e6da4cbdb1f797c48b693e91018153a/src/backend/utils/adt/float.c#L2891), feeds into [float8_combine](https://github.com/postgres/postgres/blob/f5cc81719e6da4cbdb1f797c48b693e91018153a/src/backend/utils/adt/float.c#L2951)

test/expected/partitioning_http.out
serprex: will go with both

src/deparse.c
serprex: it can throw, but yes after throw unused & memory context is freed

src/deparse.c
theory: Would you add a comment to that effect, please?

doc/pg_clickhouse.md
```diff
@@ -577,6 +577,80 @@ try=# EXPLAIN (ANALYZE, VERBOSE)
  the number of rows that must be pulled back into Postgres from 1000 (all of
  them) to just 8, one for each node.
 
+### Partitioned Tables
+
+A PostgreSQL [partitioned table] can mix local partitions with foreign
+partitions backed by ClickHouse. A common layout offloads cold data to
+ClickHouse while recent data stays in PostgreSQL:
```
theory: Mixing cold/hot with old/recent. Pick one.

doc/pg_clickhouse.md
```diff
@@ -577,6 +577,80 @@ try=# EXPLAIN (ANALYZE, VERBOSE)
...
+    FOR VALUES FROM ('2023-01-01') TO ('2024-01-01')
+    SERVER ch_svr OPTIONS (table_name 'events');
+
+-- Hot 2024 data stays local
+CREATE TABLE events_hot PARTITION OF events
+    FOR VALUES FROM ('2024-01-01') TO ('2025-01-01');
+```
+
+Relocating data from a local to a foreign partition is left as an exercise
+for the reader.
```
theory: What? Isn't that the point of this PR?

doc/pg_clickhouse.md
```diff
@@ -577,6 +577,80 @@ try=# EXPLAIN (ANALYZE, VERBOSE)
...
+
+Aggregates spanning both local and foreign partitions need [partitionwise
+aggregation], which PostgreSQL disables by default:
+
+```pgsql
+SET enable_partitionwise_aggregate = on;
+```
+
+With it, PostgreSQL computes a *partial aggregate* below `Append`, then a
```
```suggestion
With `enable_partitionwise_aggregate` enabled, PostgreSQL computes a *partial aggregate* below `Append`, then a
```

doc/pg_clickhouse.md
```diff
@@ -577,6 +577,80 @@ try=# EXPLAIN (ANALYZE, VERBOSE)
...
+               Remote SQL: SELECT count(*), sum(val), min(ts), max(ts) FROM events
+         ->  Partial Aggregate
+               ->  Seq Scan on events_hot
+```
+
+#### When partial aggregates push down
+
+PostgreSQL represents a partial aggregate as a *transition state* that the
+finalize step combines across partitions. pg_clickhouse can push a partition's
+partial down only when it can express that as a ClickHouse value:
```
```suggestion
partial down only when it can express it as a ClickHouse value:
```

doc/pg_clickhouse.md
```diff
@@ -577,6 +577,80 @@ try=# EXPLAIN (ANALYZE, VERBOSE)
...
+
+*   **Decomposable aggregates** whose transition state is already the final
+    value push down directly: `count`, `sum`, `min`, `max`, `bool_and`/`every`,
```
```suggestion
    value, push down directly: `count`, `sum`, `min`, `max`, `bool_and`/`every`,
```

doc/pg_clickhouse.md
```diff
@@ -577,6 +577,80 @@ try=# EXPLAIN (ANALYZE, VERBOSE)
...
+    `bool_or`, `bit_and`, `bit_or`, and `bit_xor`.
+*   **`avg` over integers** pushes its `{count, sum}` state as an array.
+*   **`avg`, `var_pop`, `var_samp`, `stddev_pop`, and `stddev_samp` over
+    floating point** push their `{N, sum, sum of squared deviations}` state as
+    an array.
+
+`FILTER (WHERE …)` pushes down with these.
```
```suggestion
`FILTER (WHERE …)` pushes down with these aggregate functions.
```

theory (changes requested):

doc/pg_clickhouse.md
serprex: No, this PR enables query planning once mixed partitioning is set up, but my example stored procedure to offload partitions is left out. This way we can iterate on production workflows before trying to codify anything

theory: Okay, but maybe omit this line, which some will find passive-aggressive.

CHANGELOG.md
```diff
@@ -43,6 +43,15 @@ All notable changes to this project will be documented in this file. It uses the
...
     `btrim` ([#307]).
+*   Added partial aggregates under partitionwise aggregation, so a query
+    over a partitioned table mixing local and foreign partitions computes the
+    foreign partition's aggregate on ClickHouse instead of fetching its rows.
+    Covers decomposable aggregates (`count`, `sum`, `min`, `max`,
+    `bool_and`/`bool_or`, `bit_and`/`bit_or`), plus `avg` over integers and
+    `avg`/`var_pop`/`var_samp`/`stddev_pop`/`stddev_samp` over floats.
+    Aggregates with an internal transition state (anything over `numeric`,
+    `avg(bigint)`, `avg(interval)`) still fall back to local aggregation.
+    Requires `enable_partitionwise_aggregate` ([#298]).
```
```suggestion
*   Added pushdown support for partial aggregates under partitionwise
    aggregation, so a query over a partitioned table mixing local and foreign
    partitions computes the foreign partition's aggregate on ClickHouse
    instead of fetching its rows. Covers decomposable aggregates (`count`,
    `sum`, `min`, `max`, `bool_and`/`bool_or`, `bit_and`/`bit_or`), plus `avg`
    over integers and `avg`/`var_pop`/`var_samp`/`stddev_pop`/`stddev_samp`
    over floats. Aggregates with an internal transition state (anything over
    `numeric`, `avg(bigint)`, `avg(interval)`) still fall back to local
    aggregation. Requires `enable_partitionwise_aggregate` ([#298]).
```

theory (approved):

doc/pg_clickhouse.md
serprex: will add example sql that can be used to docs, but that includes generating CREATE TABLE & it'd be nice if it could take fdw server name using binary driver (& not have some hack where we extract connection string from fdw server), so I want https://github.com/ClickHouse/pg_clickhouse/pull/309 for this

PR #299, opened 2026-06-27
push down any_value

serprex:


src/custom_types.c
```diff
@@ -697,6 +697,12 @@ lookup_builtin_func(Oid funcid, builtin_func_def* def) {
         def->ch_name = "var_samp";
         return true;
 
+#if PG_VERSION_NUM >= 160000
+    case F_ANY_VALUE:
+        def->ch_name = "any";
+        return true;
+#endif
```
theory, line 709: We typically define missing constants above, this one in the `#if PG_VERSION_NUM < 160000` section.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ All notable changes to this project will be documented in this file. It uses the
     *   [stddev_samp/stddev]
     *   [var_op]
     *   [var_samp/variance]
+    *   [any_value]
```
theory, line 22: Please add to `doc/pg_clickhouse.md`.

theory (approved):
Just needs a doc item, really. Odd that I had it listed as mapped already, not sure how I missed it.

src/custom_types.c
serprex: now that we've consolidated switch so this is one ifdef, wondering if we want to clean up previous define-missing-oids. there's two cases to tease out: OIDs that exist on that pg version but don't have constant, & OIDs that don't exist on that pg version

PR #300, opened 2026-06-27
fix(http): handle subsecond precision for inserts

### LINKED ISSUE #295: DateTime64 milliseconds are truncated when inserting via pg_clickhouse
### Description

When inserting into a `DateTime64(3)` column via pg_clickhouse, milliseconds are silently truncated. The same insert executed directly against ClickHouse stores the value correctly.

### Reproduction

**Schema**
```sql
CREATE TABLE warehouse.test (
    ts DateTime64(3) NOT NULL CODEC(DoubleDelta, ZSTD(1))
)
ENGINE = MergeTree()
PARTITION BY toYYYYMM(ts)
ORDER BY (ts);
```

**Insert directly via ClickHouse**
```sql
INSERT INTO warehouse.test (ts) VALUES ('2026-06-08 08:45:17.345');
```

**Insert via pg_clickhouse**
```sql
INSERT INTO warehouse.test (ts) VALUES ('2026-06-08 08:45:18.345');
```

**Result**
```
2026-06-08 08:45:17.345   ← direct ClickHouse insert ✅
2026-06-08 08:45:18.000   ← pg_clickhouse insert ❌
```

### Root Cause (suspected)

The pg_clickhouse type mapping table does not include `DateTime64`. It appears the layer falls back to the `DateTime` → `timestamptz` mapping, which is seconds-precision only, causing the sub-second component to be dropped before the value is written to the `DateTime64(3)` column.

### Expected Behavior

Milliseconds (and sub-second precision generally) should be preserved when inserting into a `DateTime64(N)` column via pg_clickhouse, consistent with direct ClickHouse inserts.

serprex:
Comment about ClickHouse support predated DateTime64

fixes #295


src/deparse.c
```diff
@@ -2375,10 +2374,6 @@ deparseConst(Const* node, deparse_expr_cxt* context, int showtype) {
     getTypeOutputInfo(node->consttype, &typoutput, &typIsVarlena);
 
     if (typoutput == F_TIMESTAMPTZ_OUT || typoutput == F_TIMESTAMP_OUT) {
-        /*
-         * We use our own function here, that removes fractional seconds since
```
theory, line 2379: Did we already remove this function?

theory (approved):

src/deparse.c
serprex: `ch_timestamp_out` is the function, there's more than fractional seconds to remove, it makes sure format is to ClickHouse's liking (DateStyle woes)

PR #303, opened 2026-06-28
binary driver autoflush

serprex:
binary driver was batching insert or `COPY FROM` in memory, extend so that after 64MiB it flushes. Keeps memory usage streaming while ClickHouse combines flushed blocks on its end within an insert

`ExecForeignBatchInsert` not needed for pg_clickhouse as buffering/flushing is already implemented

### DIFF
--- a/doc/pg_clickhouse.md
+++ b/doc/pg_clickhouse.md
@@ -679,11 +679,15 @@ try=# COPY logs FROM stdin CSV;
 >> COPY 3
 ```
 
-> **⚠️ Batch API Limitations**
->
-> pg_clickhouse has not yet implemented support for the PostgreSQL FDW batch
-> insert API. Thus [COPY] currently uses [INSERT](#insert) statements to
-> insert records. This will be improved in a future release.
+`COPY FROM` streams the rows to ClickHouse in bulk.
+
+`COPY TO` directly from a foreign table is not supported: PostgreSQL itself
+rejects `COPY <foreign_table> TO` with the hint to use the `COPY (SELECT ...) TO`
+variant. Use that variant to copy rows out of a ClickHouse foreign table:
+
+```pgsql
+try=# COPY (SELECT * FROM logs) TO stdout CSV;
+```
 
 ### LOAD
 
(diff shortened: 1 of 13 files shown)


doc/pg_clickhouse.md
```diff
@@ -679,11 +679,15 @@ try=# COPY logs FROM stdin CSV;
 >> COPY 3
 ```
 
-> **⚠️ Batch API Limitations**
->
-> pg_clickhouse has not yet implemented support for the PostgreSQL FDW batch
-> insert API. Thus [COPY] currently uses [INSERT](#insert) statements to
-> insert records. This will be improved in a future release.
+`COPY FROM` streams the rows to ClickHouse in bulk.
```
theory, line 682: Let's add a comment to `clickhouse_fdw_handler()` explaining that `ExecForeignBatchInsert` and `GetForeignModifyBatchSize` aren't necessary thanks to ClickHouse's streaming of `INSERT`s. Remove or update the comment starting `TODO:Add support for ClickHouse 25.8 and later.` (line 3456).

Though I presume this doesn't apply to the http driver, yes? If so, let's say as much.

src/binary/insert.c
```diff
@@ -1239,6 +1239,28 @@ build_lc_dict(
...
+ch_binary_insert_autoflush(ch_binary_insert_state* state) {
+    ch_binary_insert_handle* h = state->insert_block;
+
+    if (h) {
+        size_t total = 0;
+        for (size_t i = 0; i < h->ncols; i++) {
+            const ic_col* c = &h->cols[i];
+            total += c->body.len + c->nulls.len +
+                     (c->body_offs.len + c->arr_offs.len) * sizeof(uint64_t);
+        }
```
theory: Should we have the connection object track this per insert?

theory (approved):

doc/pg_clickhouse.md
serprex: http driver already has its own batching

src/binary/insert.c
serprex: no, tried it out, we end up having some counter we now need to keep on heap & it still needs to be accumulated into per row

PR #307, opened 2026-06-30
Push down three-argument trim functions

theory:
`btrim` was already supported, but untested. Add `ltrim` and `rtrim` and a test for them. Also fix the links to the ClickHouse docs for `ltrim` and `btrim`.


serprex (approved):

theory:
> description means rtrim when it says btrim?

Well the order could be better. Changed in the final commit.

PR #309, opened 2026-06-30
add binary support to clickhouse_raw_query, add clickhouse_query

### LINKED ISSUE #30: Support `UPDATE`, `DELETE`
Available in ClickHouse 25.8 and higher. Might require changes to [clickhouse-cpp](/ClickHouse/clickhouse-cpp).

### LINKED ISSUE #80: Add Direct Query Function
`clickhouse_raw_query()` currently collects and displays all the results of a query as a single TEXT value. Determine whether it'd be feasible to provide a similar function that takes a native ClickHouse query but maps the return values to Postgres data types and column names and displays them like table, as table-returning functions to. This similar to how [JSON_TABLE](https://www.postgresql.org/docs/current/functions-json.html#FUNCTIONS-SQLJSON-TABLE) works; see if we can borrow the code there to dynamically determine column names and types.

serprex:
fixes #30
fixes #80

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,10 @@ All notable changes to this project will be documented in this file. It uses the
     bounding memory for large `COPY FROM` and `INSERT SELECT` ([#303]).
 *   Added pushdown for the three-argument forms of `ltrim`, `rtrim`, and
     `btrim` ([#307]).
+*   Added support for binary driver to `clickhouse_raw_query()`.
+*   Added `clickhouse_query(server, sql)`, a set-returning function that runs a
+    query against a configured foreign server and returns its rows typed by the
+    caller's column definition list.
 
 ### 🐞 Bug Fixes
 
--- a/doc/pg_clickhouse.md
+++ b/doc/pg_clickhouse.md
@@ -1006,17 +1006,24 @@ SELECT clickhouse_raw_query(
 );
 ```
 
-Connect to a ClickHouse service via its HTTP interface, execute a single
-query, and disconnect. The optional second argument specifies a connection
-string that defaults to `host=localhost port=8123`. The supported connection
-parameters are:
+Connect to a ClickHouse service, execute a single query, and disconnect. The
+optional second argument specifies a connection string that defaults to
+`host=localhost port=8123`. The supported connection parameters are:
 
+*   `driver`: The connection driver to use, either "http" or "binary"; defaults
+    to "http"
 *   `host`: The host to connect to; required.
-*   `port`: The HTTP port to connect to; defaults to `8123` unless `host` is a
-    ClickHouse Cloud host, in which case it defaults to `8443`
+*   `port`: The port to connect to. Defaults to `8123` for the "http" driver or
+    `9000` for the "binary" driver, switching to `8443` or `9440` respectively
+    when `host` is a ClickHouse Cloud host
 *   `dbname`: The name of the database to connect to.
 *   `username`: The username to connect as; defaults to `default`
 *   `password`: The password to use to authenticate; defaults to no password
+
+Both drivers return tab-separated rows (nulls as `\N`), but their per-value
+representations differ: the "http" driver returns ClickHouse's own TSV
+formatting verbatim, whereas the "binary" driver passes each value through its
+PostgreSQL output function.
 
 By default, no role has `EXECUTE` access to this function; consider [GRANT]ing
 access only to roles that legitimately need to execute ad-hoc ClickHouse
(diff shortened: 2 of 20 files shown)


doc/pg_clickhouse.md
```diff
@@ -1006,18 +1006,25 @@ SELECT clickhouse_raw_query(
...
+    `9000` for the "binary" driver, switching to `8443` or `9440` respectively
+    when `host` is a ClickHouse Cloud host
 *   `dbname`: The name of the database to connect to.
 *   `username`: The username to connect as; defaults to `default`
 *   `password`: The password to use to authenticate; defaults to no password
 
+The "binary" driver renders each result row as tab-separated column values
+(nulls as `\N`), passing each value through its PostgreSQL output function. Its
+formatting therefore differs from the "http" driver, which returns ClickHouse's
+own wire formatting verbatim.
```
theory: Isn't that also TSVs?

doc/pg_clickhouse.md
```diff
@@ -1068,6 +1075,33 @@ Reads the version from the native-protocol connection handshake, or over HTTP
...
+rows as a relation, mapping each ClickHouse result column to the PostgreSQL type
+named in the column definition list. Unlike [`clickhouse_raw_query`](#clickhouse_raw_query)
+it reuses the server's `driver`, credentials, database, and the connection cache
+rather than an ad-hoc connection string.
+
+The first argument is the name of a server created with [CREATE SERVER]. A
+column definition list (`AS name(col type, ...)`) is required: PostgreSQL needs
+the result shape before fetching rows, and it must match the columns the query
+returns. Values are converted from ClickHouse to the declared types the same way
+a foreign table column would be.
```
theory, line 1097: I like this. Wonder if we should also support something like the `pgduckdb.row` type for queries without a column spec (would have to be a different function, presumably).

doc/pg_clickhouse.md
```diff
@@ -1068,6 +1075,33 @@ Reads the version from the native-protocol connection handshake, or over HTTP
...
+
+As with `clickhouse_raw_query`, no role has `EXECUTE` access by default:
```
```suggestion
As with `clickhouse_raw_query`, no role has `EXECUTE` access by default; 
`GRANT` to a role to allow it to use the function.
```

sql/pg_clickhouse.sql
theory, line 1: Need an upgrade script with the same statement.

src/fdw.c
```diff
@@ -408,6 +426,112 @@ clickhouse_server_version(PG_FUNCTION_ARGS) {
...
+    }
+
+    /* The result shape comes from the caller's column definition list. */
+    if (get_call_result_type(fcinfo, NULL, &tupdesc) != TYPEFUNC_COMPOSITE) {
+        ereport(
+            ERROR,
+            (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+             errmsg("a column definition list is required for clickhouse_query"))
+        );
+    }
```
theory, line 475: So if it's missing, could we not use the info returned by the driver to populate them? (Future feature)

src/fdw.c
```diff
@@ -408,6 +426,112 @@ clickhouse_server_version(PG_FUNCTION_ARGS) {
...
+        MemoryContextReset(row_cxt);
+        CHECK_FOR_INTERRUPTS();
+    }
+
+    MemoryContextDelete(row_cxt);
+    MemoryContextDelete(cursor->memcxt);
+    chfdw_release_scan_connection(user, sconn);
+
+    return (Datum)0;
+}
```
theory, line 541: I would appreciate comments describing what each block of code is doing. Where does it set names and types, for example?

src/pglink.c
```diff
@@ -1005,6 +1005,94 @@ binary_simple_query(void* conn, const ch_query* query) {
...
+        case '\r':
+            appendStringInfoString(buf, "\\r");
+            break;
+        case '\t':
+            appendStringInfoString(buf, "\\t");
+            break;
+        default:
+            appendStringInfoChar(buf, *s);
+        }
+    }
```
theory, line 1038: How many copies of this do we have now 🫠

CHANGELOG.md
```diff
@@ -43,6 +43,10 @@ All notable changes to this project will be documented in this file. It uses the
     bounding memory for large `COPY FROM` and `INSERT SELECT` ([#303]).
 *   Added pushdown for the three-argument forms of `ltrim`, `rtrim`, and
     `btrim` ([#307]).
+*   Added support for binary driver in `clickhouse_raw_query()`.
```
```suggestion
*   Added support for binary driver to `clickhouse_raw_query()`.
```

theory (approved):
Love it. A few nits.

doc/pg_clickhouse.md
serprex: Yes, but not the same value representation

doc/pg_clickhouse.md
serprex: That was the other option, would be a much larger lift, & has its own drawbacks having to gather everything into some pseudo-row value that then has to be unpacked at some later point

src/pglink.c
serprex: main qualm with `ch_escape_string` is that it escapes `'` while http result doesn't. could add a `bool escape_quotes` arg

needing to shoehorn binary result into `text` is a strong reason to deprecate `clickhouse_raw_query`

src/fdw.c
serprex: No, need different function, since pg function is returning `record` which needs to have explicit typing by time we get here

doc/pg_clickhouse.md
theory: I like this better, TBH; it relies on the exiting Postgres interface.

src/pglink.c
theory: Yeah, I'm all for deprecating it, was always ugly. This is _much_ nicer. Though it's nice to also support queries without an existing server. But maybe no one will complain, and we're better off security-wise relying on user mappings.

PR #310, opened 2026-07-01
Disable MinMaxAgg without impacting overall query planner costs

### LINKED ISSUE #130: Increase MIN/MAX optimization path estimation
In 02bf035 (#16), `estimate_path_cost_size()` was tweaked to estimate a lower cost so as to prevent the MIN/MAX optimization that Postgres creates from taking priority. This works well-enough, but it might be better to instead search through the paths in the root node, find any remote `min/max` optimization paths, and increase their costs, to convince the planner not to use them. This seems more honest, since we don't want the optimization to be pushed down to ClickHouse, and would allow more accurate costing to be implemented in `estimate_path_cost_size()`.

serprex:
fixes #130

### DIFF
--- a/src/fdw.c
+++ b/src/fdw.c
@@ -1799,7 +1799,7 @@ estimate_path_cost_size(
     *p_rows         = 1000;
     *p_width        = 32;
     *p_startup_cost = 1.0;
-    *p_total_cost   = 0.1 + coef;
+    *p_total_cost   = 5.0 + coef;
 }
 
 /*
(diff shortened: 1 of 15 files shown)


src/fdw.c
```diff
@@ -3009,6 +3010,19 @@ add_foreign_grouping_paths(
     ); /* no fdw_private */
 #endif
 
+    /*
+     * For simple min/max queries Postgres adds MinMaxAggPath that rewrites as
+     * ORDER BY col LIMIT 1. LIMIT invisible to FDW, causing full remote scan.
```
theory:
```suggestion
     * For simple min/max queries Postgres adds MinMaxAggPath that rewrites as
     * ORDER BY col LIMIT 1. LIMIT is invisible to FDW, causing full remote
     * scan, so increase the cost to prevent this optimization.
```

theory (approved):
Please add CHANGELOG item.

PR #313, opened 2026-07-10
Fix misc undefined behavior

serprex:
1. kv_iter_next would access out of bounds with strlen when iter->togo == 1
2. CollapsingMergeTree was didn't validate well with bad parenthesis
3. CF_DATE_TRUNC/CF_DATE_PART didn't reject non-Const or null units
4. binary_simple_query was accessing state->coltypes[j] potentially out of bounds if column counts mismatched
5. volatile MemoryContext tempcxt needed for longjmp/setjmp semantics, added -Wclobbered to warn about this on gcc
6. char_to_datum would skip out of bounds on inputs without prefix
7. avoid calling convert_record / ch_binary_init_convert_state with NULL
8. `busy` should also be initialized in get_connection_entry
9. use typed curl_xferinfo_callback
10. curl_easy_escape returns NULL on failure, handle failure
11. DateTime64 scale bounded against lengthof(pow10i) for more hardening against malicious CH server
12. Wsign-compare

### DIFF
--- a/src/custom_types.c
+++ b/src/custom_types.c
@@ -1086,18 +1086,28 @@ chfdw_apply_custom_table_options(CHFdwRelationInfo* fpinfo, Oid relid) {
             if (strncasecmp(val, collapsing_text, strlen(collapsing_text)) == 0) {
                 char *start = index(val, '('), *end = rindex(val, ')');
                 char sign[CH_ESCAPED_NAMEDATALEN];
+                Size sign_len;
 
-                if (end - start - 1 > (CH_ESCAPED_NAMEDATALEN - 1)) {
+                if (start == NULL || end == NULL || end <= start + 1) {
                     ereport(
                         ERROR,
-                        (errcode(ERRCODE_FDW_INVALID_ATTRIBUTE_VALUE),
-                         errmsg("pg_clickhouse: invalid engine parameter"))
+                        errcode(ERRCODE_FDW_INVALID_ATTRIBUTE_VALUE),
+                        errmsg("pg_clickhouse: invalid engine parameter")
+                    );
+                }
+
+                sign_len = end - start - 1;
+                if (sign_len > CH_ESCAPED_NAMEDATALEN - 1) {
+                    ereport(
+                        ERROR,
+                        errcode(ERRCODE_FDW_INVALID_ATTRIBUTE_VALUE),
+                        errmsg("pg_clickhouse: invalid engine parameter")
                     );
                 }
 
                 fpinfo->ch_table_engine = CH_COLLAPSING_MERGE_TREE;
-                strncpy(sign, start + 1, end - start - 1);
-                sign[end - start - 1] = '\0';
+                memcpy(sign, start + 1, sign_len);
+                sign[sign_len] = '\0';
                 strlcpy(
                     fpinfo->ch_table_sign_field,
                     ch_quote_ident(sign),
--- a/src/include/kv_list.h
+++ b/src/include/kv_list.h
@@ -18,15 +18,19 @@ typedef struct kv_list {
 } kv_list;
 
 /*
- * Iterator for a kv_list. Use new_kv_iter() to create.
+ * Iterator for a kv_list. Use new_kv_iter() to create; name and value are
+ * valid only after kv_iter_next() returns true.
  *
- *     for (kv_iter iter = new_kv_iter(kv); !kv_iter_done(&iter); kv_iter_next(&iter))
+ *     kv_iter iter = new_kv_iter(kv);
+ *     while (kv_iter_next(&iter))
  *     {
- *         printf("%i, %s => %s\n", iter.num, iter.name, iter.value);
+ *         printf("%s => %s\n", iter.name, iter.value);
  *     }
  */
 typedef struct kv_iter {
     int togo;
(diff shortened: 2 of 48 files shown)


test/expected/binary.out
```diff
@@ -438,13 +438,68 @@ LINE 1: SELECT * FROM clickhouse_query('binary_loopback', 'SELECT 1'...
...
+    SELECT tuple(
+        if(number = 0,
+           CAST(NULL, 'Nullable(Tuple(Int32))'),
+           CAST(tuple(toInt32(number)), 'Nullable(Tuple(Int32))'))
+    )
+    FROM numbers(2)
+    ORDER BY number
+    SETTINGS allow_experimental_nullable_tuple_type = 1
+    $$
+) AS t(v text);
```
theory, line 488: Presumably works with a composite type.

test/expected/result_map.txt
theory, line 1: Need to add section for binary.sql now, yes?

test/expected/functions_0.out
theory, line 1: Oof, we're going to have to split up this file.

test/expected/result_map.txt
```diff
@@ -109,8 +109,9 @@ functions.sql
 ----------|---------------
  19       | functions.out
  18       | functions_1.out
- 14-17    | functions_2.out
+ 15-17    | functions_2.out
  13       | functions_3.out
+ 14       | functions_0.out
```
theory: reverse order of these lines

src/include/kv_list.h
```diff
@@ -18,15 +18,19 @@ typedef struct kv_list {
 } kv_list;
 
 /*
- * Iterator for a kv_list. Use new_kv_iter() to create.
+ * Iterator for a kv_list. Use new_kv_iter() to create. name and value are
```
theory:
```suggestion
 * Iterator for a kv_list. Use new_kv_iter() to create; name and value are
```

src/custom_types.c
```diff
@@ -1086,8 +1086,18 @@ chfdw_apply_custom_table_options(CHFdwRelationInfo* fpinfo, Oid relid) {
...
                 char *start = index(val, '('), *end = rindex(val, ')');
                 char sign[CH_ESCAPED_NAMEDATALEN];
+                Size sign_len;
 
-                if (end - start - 1 > (CH_ESCAPED_NAMEDATALEN - 1)) {
+                if (start == NULL || end == NULL || end <= start + 1) {
+                    ereport(
+                        ERROR,
+                        (errcode(ERRCODE_FDW_INVALID_ATTRIBUTE_VALUE),
+                         errmsg("pg_clickhouse: invalid engine parameter"))
```
theory:
```suggestion
                        errcode(ERRCODE_FDW_INVALID_ATTRIBUTE_VALUE),
                        errmsg("pg_clickhouse: invalid engine parameter")
```

src/pglink.c
```diff
@@ -307,8 +307,10 @@ report_http_stream_query_failure(
 
 static ch_cursor*
 http_simple_query(void* conn, const ch_query* query) {
-    int attempts          = 0;
-    MemoryContext tempcxt = NULL, oldcxt;
+    int attempts = 0;
+    /* volatile: assigned between PG_TRY setjmp and PG_CATCH read */
```
theory: What does this prevent/fix?

theory (approved):

src/pglink.c
serprex: [`longjmp` ](https://gcc.gnu.org/onlinedocs/gcc-8.5.0/gcc/Global-Register-Variables.html)

> On some machines, however, longjmp does not change the value of global register variables. To be portable, the function that called setjmp should make other arrangements to save the values of the global register variables, and to restore them in a longjmp.

We could decide to trust variables in registers get preserved, but this is necessary to be spec compliant

theory: Would you add a comment to this effect?

src/custom_types.c
serprex: fixed them all

test/expected/binary.out
serprex: yes, added

src/custom_types.c
theory: lol, thanks!

test/expected/binary.out
```diff
@@ -438,13 +438,104 @@ LINE 1: SELECT * FROM clickhouse_query('binary_loopback', 'SELECT 1'...
...
+    )
+    FROM numbers(2)
+    ORDER BY number
+    SETTINGS allow_experimental_nullable_tuple_type = 1
+    $$
+) AS t(v nested_tup);
+    v    
+---------
+ ()
+ ("(1)")
```
theory, line 534: That doesn't look right. Should be:

```
 ((1))
```

No?

test/expected/binary.out
theory: Should be `((1))`, not `("(1)")`, no?

test/expected/binary.out
serprex: no,
```
> select ROW(1),ROW(ROW(1));
 row |   row   
-----+---------
 (1) | ("(1)")
```

PR #315, opened 2026-07-13
Preserve NULL semantics in IN-family pushdown

JoshDreamland:
ClickHouse evaluates IN under binary Boolean logic where the Postgres
expression can result in NULL, so pushed `IN`-family expressions gave
wrong answers whenever a NULL could reach the comparison:
`x NOT IN (1, NULL)` returned rows Postgres would drop, meanwhile
`has()`/`countEqual()` matched a NULL probe against NULL array elements
as values, `x NOT IN (SELECT ...)` over nullable columns returned the
plain set complement, and grouping by an `IN` result merged Postgres's
`NULL` group into `FALSE`. The same audit caught three walker bugs:
1. `<> ANY(array)` deparsed to SQL computing `<> ALL`
   (wrong with no NULLs involved),
2. `CASE arg WHEN ..` shipped unchecked when the tested expression was
   unshippable, and
3. a bare EXPLAIN of `x = ANY(NULL::int[])` crashed the backend when
   detoasting the null array datum.

To see each of these, run the new `in_null_semantics` test and
`subplan_pushdown` section 10 against `main`: the wrong rows appear as
regression diffs and the `EXPLAIN` as the crash.

The shippability walker now tracks how each expression's result is
consumed, which I'm generally apprehensive of doing, but it avoids
blowing up this PR adding specialized handlers for different contexts.
Filter conditions treat `NULL` exactly like `FALSE`, so ClickHouse's
`FALSE`-for-`NULL` substitution is invisible there and those shapes
ship unchanged. Value positions and negations demand exact three-valued
equivalence, established by proving the operands can never be `NULL`
(non-NULL constants; NOT NULL columns not nulled by outer joins).
`x NOT IN (SELECT ...)` filters keep their pushdown either way:
a clean query when the columns prove non-nullable, a guarded one
enforcing Postgres's answer when they do not, each guard emitted only
when its proof (of non-nullability) fails. Grouped or aggregated
subquery bodies without proofs stay local, as do `<> ANY` and
`NULL`-array shapes, which will require a different spelling as well
that I think deserves a follow-up PR (modulo the pathological case of
someone attempting to construct a NULL array as part of the query).

`in_null_semantics` carries one expected file across PG 13-19;
subplan_pushdown gains a nullable `NOT IN` section on all four run
variants.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,18 @@ All notable changes to this project will be documented in this file. It uses the
 *   Added the `clickhouse_server_version(server)` function, which reports the
     ClickHouse server version ("major.minor.patch") for a foreign server
     ([#293]).
+*   Added pushdown for subqueries the planner cannot flatten into joins
+    (SubPlans): correlated and uncorrelated scalar aggregate subqueries,
+    `EXISTS`, and single-column equality `IN`/`NOT IN`, in `WHERE` and `HAVING`.
+    Covers TPC-H Q2, Q11, Q15, Q17, Q20, and Q22 shapes. `NOT IN` preserves
+    PostgreSQL's NULL semantics under ClickHouse's two-valued `IN`, deparsing
+    with compensating guards when the columns involved are nullable ([#315]).
+    Requires ClickHouse 25.8 or later; older servers keep evaluating the
+    subquery locally ([#289]).
+*   Added `transform_null_in 0` to the default value of the
+    `pg_clickhouse.session_settings` parameter, so that a ClickHouse server
+    profile cannot silently change the `IN` semantics the pushdown rules rely
+    on ([#315]).
 *   Added pushdown for more aggregate functions ([#290]):
     *   [corr]
     *   [covarpop]
@@ -74,6 +86,16 @@ All notable changes to this project will be documented in this file. It uses the
     rejection of non-`Const`/null units in `date_trunc`/`date_part` pushdown;
     NULL handling in record conversion; unchecked `curl_easy_escape` failures;
     and unbounded `DateTime64` scale lookups ([#313]).
+*   Fixed wrong results from pushed-down `= ANY`/`<> ALL` expressions and `IN`
+    expressions over constant lists and when a `NULL` could reach the
+    comparison, because ClickHouse evaluates `IN` under two-valued logic while
+    PostgreSQL evaluates three-value `NULL` logic. These cases now ship only
+    where PostgreSQL's three-valued answer can be provably preserved ([#315]).
+*   Fixed `<> ANY(array)` deparsing to ClickHouse SQL that computes `<> ALL`,
+    which is wrong even with no `NULL`s involved: In Postgres,
+    `1 <> ANY('{1,5}')` is `TRUE`. For now, do not push down ([#315]).
+*   Fixed the pushdown of a `CASE arg WHEN ..` expression without checking its
+    branches when the tested expression was itself unshippable ([#315]).
 
 ### 📚 Documentation
 
@@ -114,8 +136,12 @@ All notable changes to this project will be documented in this file. It uses the
     "ClickHouse/pg_clickhouse#309 add binary support to clickhouse_raw_query, add clickhouse_query"
   [#310]: https://github.com/ClickHouse/pg_clickhouse/pull/310
     "ClickHouse/pg_clickhouse#310 Disable MinMaxAgg without impacting overall query planner costs"
+  [#289]: https://github.com/ClickHouse/pg_clickhouse/pull/289
+    "ClickHouse/pg_clickhouse#289 Push down subquery (SubPlan) residue to ClickHouse"
   [#313]: https://github.com/ClickHouse/pg_clickhouse/pull/313
     "ClickHouse/pg_clickhouse#313 Fix misc undefined behavior"
+  [#315]: https://github.com/ClickHouse/pg_clickhouse/pull/315
+    "ClickHouse/pg_clickhouse#315 Preserve NULL semantics in IN-family pushdown"
 
 ## [v0.3.2] — 2026-06-16
 
--- a/doc/pg_clickhouse.md
+++ b/doc/pg_clickhouse.md
@@ -795,8 +795,12 @@ settings] to be set on subsequent queries. Example:
 SET pg_clickhouse.session_settings = 'join_use_nulls 1, final 1';
 ```
 
-The default is `join_use_nulls 1, group_by_use_nulls 1, final 1`. Set it to an
-empty string to fall back on the ClickHouse server's settings.
+The default is `join_use_nulls 1, group_by_use_nulls 1, final 1,
+transform_null_in 0`. Set it to an empty string to fall back on the
+ClickHouse server's settings — but note that pushdown correctness depends on
+some of these defaults: `join_use_nulls` for outer joins and
+`transform_null_in` for the `IN` family (see [IN and NULL
+Semantics](#in-and-null-semantics)).
 
 ```sql
 SET pg_clickhouse.session_settings = '';
@@ -1309,6 +1313,30 @@ equivalents as follows:
 *   `!~*` (case insensitive regexp not match): [match](https://clickhouse.com/docs/sql-reference/functions/string-search-functions#match)
 *   `->>` (JSON/JSONB extract element as text): [sub-column syntax](https://clickhouse.com/docs/sql-reference/data-types/newjson#reading-json-paths-as-sub-columns)
 *   `->` (JSON/JSONB extract): [toJSONString](https://clickhouse.com/docs/sql-reference/functions/json-functions#toJSONString) + [sub-column syntax](https://clickhouse.com/docs/sql-reference/data-types/newjson#reading-json-paths-as-sub-columns)
+
+### IN and NULL Semantics
+
+ClickHouse evaluates `IN` under two-valued logic: when the probe finds no
+match it returns `0` even if a NULL is involved, where PostgreSQL computes
+NULL. To preserve PostgreSQL semantics, pg_clickhouse pushes down the `IN`
+family (`IN`, `NOT IN`, `= ANY`, `<> ALL`, and `IN (SELECT ...)`) only where
+the two systems provably agree: in a plain filter condition, where a NULL
+disqualifies a row exactly like `FALSE`, or when none of the probe, the list,
+nor the subquery output can produce a NULL. A `NOT IN (SELECT ...)` filter
+over nullable columns still pushes down, deparsed with compensating guards
+that keep PostgreSQL's answer: a set containing a NULL disqualifies every
+row, and a NULL probe passes only against an empty set. Each guard is
+omitted when a `NOT NULL` declaration proves it unnecessary. The remaining
+shapes (NULL-capable `IN` expressions in select lists, `GROUP BY`, or
+`ORDER BY`, and `NOT IN` over arrays or grouped subqueries) are evaluated
+locally. Declaring columns `NOT NULL` maximizes pushdown; [IMPORT FOREIGN
+SCHEMA] does so automatically for non-`Nullable` ClickHouse columns.
+
+These rules assume ClickHouse's default `transform_null_in = 0`, which
+pg_clickhouse sets on every query through the default value of the
+[`pg_clickhouse.session_settings`](#pg_clickhousesession_settings) parameter
+so that a ClickHouse server profile cannot silently change it. Setting
+`transform_null_in = 1` breaks the semantics of every pushed `IN`.
 
 ### Custom Functions
 
--- a/src/deparse.c
+++ b/src/deparse.c
@@ -173,13 +173,33 @@ typedef struct deparse_expr_cxt {
     } while (0)
 
 /*
+ * How an expression's result is consumed, for the IN-family NULL-semantics
+ * rules (see the block comment for saop_null_semantics_ok).
+ *
+ * EXPR_CTX_TRUTH    only qualifies rows: NULL and FALSE are interchangeable.
+ * EXPR_CTX_NEGATED  the direct operand of one NOT whose own result is in a
+ *                   truth context. Only a SubPlan consumes this state (it
+ *                   admits the guarded NOT IN deparse); every other node
+ *                   degrades it to EXPR_CTX_EXACT for its children, so the
+ *                   walker and the deparse peek in deparseBoolExpr agree on
+ *                   exactly which SubPlans get the guarded form.
+ * EXPR_CTX_EXACT    the value itself is observable: exact three-valued
+ *                   equivalence required.
+ */
+typedef enum ExprTruthCtx {
+    EXPR_CTX_TRUTH,
+    EXPR_CTX_NEGATED,
+    EXPR_CTX_EXACT,
+} ExprTruthCtx;
+
+/*
  * Functions to determine whether an expression can be evaluated safely on
  * remote server.
  */
 static bool
-foreign_expr_walker(Node* node, foreign_glob_cxt* glob_cxt);
+foreign_expr_walker(Node* node, foreign_glob_cxt* glob_cxt, ExprTruthCtx ctx);
 static bool
-is_shippable_subplan(SubPlan* subplan, foreign_glob_cxt* glob_cxt);
+is_shippable_subplan(SubPlan* subplan, foreign_glob_cxt* glob_cxt, ExprTruthCtx ctx);
 static char*
 deparse_type_name(Oid type_oid, int32 typemod);
 
@@ -322,10 +342,21 @@ static void
 deparseNullIfExpr(NullIfExpr* node, deparse_expr_cxt* context);
 static void
 appendRegex(List* args, deparse_expr_cxt* context);
+/*
+ * Which rendering of a SubPlan's subquery body deparseSubPlanQuery emits.
+ */
+typedef enum SubPlanBodyForm {
+    SUBPLAN_BODY_QUERY,      /* the subquery as written */
+    SUBPLAN_BODY_NULL_PROBE, /* SELECT 1 .. AND (output IS NULL): the poison
+                              * guard of a guarded NOT IN */
+} SubPlanBodyForm;
+
 static void
 deparseSubPlan(SubPlan* node, deparse_expr_cxt* context);
 static void
-deparseSubPlanQuery(SubPlan* subplan, deparse_expr_cxt* context);
+deparseGuardedNotIn(SubPlan* subplan, deparse_expr_cxt* context);
+static void
+deparseSubPlanQuery(SubPlan* subplan, deparse_expr_cxt* context, SubPlanBodyForm form);
 static void
 deparseSubPlanQuals(Node* quals, deparse_expr_cxt* context);
 static void
@@ -368,7 +399,7 @@ chfdw_classify_conditions(
     foreach (lc, input_conds) {
         RestrictInfo* ri = lfirst_node(RestrictInfo, lc);
 
-        if (chfdw_is_foreign_expr(root, baserel, ri->clause)) {
+        if (chfdw_is_foreign_expr(root, baserel, ri->clause, false)) {
             *remote_conds = lappend(*remote_conds, ri);
         } else {
             *local_conds = lappend(*local_conds, ri);
@@ -378,9 +409,24 @@ chfdw_classify_conditions(
 
 /*
  * Returns true if given expr is safe to evaluate on the foreign server.
+ *
+ * exact_ctx determines how the expression's result is consumed: false for a
+ * WHERE/JOIN/HAVING condition, where a NULL qualifies a row exactly like
+ * FALSE; true anywhere the computed value itself is observable (target
+ * lists, grouping and ordering expressions). The distinction feeds the
+ * IN-family NULL-semantics rules; see foreign_expr_walker. Internally the
+ * walker refines this into ExprTruthCtx's three states; the third
+ * (EXPR_CTX_NEGATED) is derivable only from a NOT encountered during the
+ * walk, never assertable by a caller, so the entry point speaks the
+ * two-valued form.
  */
 bool
-chfdw_is_foreign_expr(PlannerInfo* root, RelOptInfo* baserel, Expr* expr) {
+chfdw_is_foreign_expr(
+    PlannerInfo* root,
+    RelOptInfo* baserel,
+    Expr* expr,
+    bool exact_ctx
+) {
     foreign_glob_cxt glob_cxt;
     CHFdwRelationInfo* fpinfo = (CHFdwRelationInfo*)(baserel->fdw_private);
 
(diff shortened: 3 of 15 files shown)


doc/pg_clickhouse.md
```diff
@@ -1310,6 +1310,24 @@ equivalents as follows:
...
+disqualifies a row exactly like `FALSE`, or when none of the probe, the list,
+nor the subquery output can produce a NULL. A `NOT IN (SELECT ...)` filter
+over nullable columns still pushes down, deparsed with compensating guards
+that keep PostgreSQL's answer: a set containing a NULL disqualifies every
+row, and a NULL probe passes only against an empty set. Each guard is
+omitted when a `NOT NULL` declaration proves it unnecessary. The remaining
+shapes (NULL-capable `IN` expressions in select lists, `GROUP BY`, or
+`ORDER BY`, and `NOT IN` over arrays or grouped subqueries) are evaluated
+locally. Declaring columns `NOT NULL` maximizes pushdown; `IMPORT FOREIGN
+SCHEMA` does so automatically for non-`Nullable` ClickHouse columns.
```
theory: Link to import foreign schema in this doc.

src/deparse.c
```diff
@@ -168,18 +168,38 @@ typedef struct deparse_expr_cxt {
...
     do {                                                                               \
         for (int i = 0; str[i]; i++) {                                                 \
-            str[i] = tolower((unsigned char)str[i]);                                   \
+            str[i] = tolower(str[i]);                                                  \
         }                                                                              \
     } while (0)
 
+/*
+ * How an expression's result is consumed, for the IN-family NULL-semantics
+ * rules (see the block comment above saop_null_semantics_ok).
```
theory: Use of "above" confused me since it's below.

```suggestion
 * rules (see the block comment for saop_null_semantics_ok).
```

src/deparse.c
```diff
@@ -378,9 +409,24 @@ chfdw_classify_conditions(
 
 /*
  * Returns true if given expr is safe to evaluate on the foreign server.
+ *
+ * exact_ctx says how the expression's result is consumed: false for a
```
```suggestion
 * exact_ctx determines how the expression's result is consumed: false for a
```

src/deparse.c
```diff
@@ -378,9 +409,24 @@ chfdw_classify_conditions(
...
+ * WHERE/JOIN/HAVING condition, where a NULL qualifies a row exactly like
+ * FALSE; true anywhere the computed value itself is observable (target
```
theory, line 415: I presume this is because we explicitly set [join_use_nulls](https://clickhouse.com/docs/operations/settings/settings#join_use_nulls) to match the SQL standard. Please mention that fact here if it is in fact relevant.

Perhaps ClickHouse should add a setting for SQL standard NULL behavior in `IN()`, too.

src/deparse.c
```diff
@@ -504,6 +552,342 @@ agg_partial_kind(Aggref* agg) {
...
 }
 
+static bool
+is_ok_in_expr(Expr* expr);
+
+/*
+ * NULL-semantics guard for the IN family (ScalarArrayOpExpr and SubPlan ANY).
+ *
+ * ClickHouse evaluates IN under two-valued logic (its default
+ * transform_null_in = 0; the connection does not override it): a probe that
```
theory: Should we enforce `transform_null_in = 0` so that the user can't override it?

src/deparse.c
```diff
@@ -504,6 +552,342 @@ agg_partial_kind(Aggref* agg) {
...
+ * branches, and aggregate FILTERs, composed through any number of AND/OR.
+ * The walker tracks that property as exact_ctx = false ("truth context").
+ * Under NOT, in a function/operator/aggregate argument, in a grouping or
+ * ordering expression — anywhere the boolean's value is consumed — the
+ * substitution is observable, so exact_ctx = true demands exact three-valued
+ * equivalence. A TRUE-for-NULL substitution is wrong in every context, so
+ * the complemented forms require a provably NULL-free right-hand side.
+ */
+
+/*
```
theory: Odd to have two comments in a row.

src/deparse.c
```diff
@@ -504,6 +552,342 @@ agg_partial_kind(Aggref* agg) {
...
+ * True if `expr` provably cannot evaluate to NULL: a non-NULL constant, or a
+ * column of a scanned relation that is declared NOT NULL and cannot have been
+ * null-extended by an outer join. Anything else is assumed nullable.
+ *
+ * PG 17+ precomputes the catalog half of this as
+ * RelOptInfo->notnullattnums; the syscache lookup below keeps one code path
```
theory, line 564: I don't see this used anywhere; should we use it on 17 now?

src/deparse.c
```diff
@@ -504,6 +552,342 @@ agg_partial_kind(Aggref* agg) {
...
+        return !((Const*)expr)->constisnull;
+    }
+
+    if (IsA(expr, Var)) {
+        Var* var = (Var*)expr;
+        RangeTblEntry* rte;
+        HeapTuple tp;
+        bool notnull;
+
+        if (var->varlevelsup != 0 || var->varattno <= 0 ||
```
theory, line 585: What is this about?

src/deparse.c
```diff
@@ -504,6 +552,342 @@ agg_partial_kind(Aggref* agg) {
...
+        }
+        return SAOP_ARR_NULL_FREE;
+    }
+
+    return SAOP_ARR_UNKNOWN;
+}
+
+/*
+ * Decide whether a ScalarArrayOpExpr keeps Postgres semantics on ClickHouse,
+ * per the divergence rules in the block comment above. The decision follows
```
theory: Which block comment? They're all block comments!

src/deparse.c
```diff
@@ -504,6 +552,342 @@ agg_partial_kind(Aggref* agg) {
...
+    /*
+     * <> ANY has no correct deparse form yet: the not has() ClickHouse SQL it
+     * would produce computes <> ALL (Postgres: 1 <> ANY of {1,5} is TRUE, one
+     * element differs; not has({1,5}, 1) is FALSE). Always evaluate locally.
+     * A truth-context spelling exists for a future follow-up — TRUE exactly
+     * when a non-NULL element differs from a non-NULL probe:
+     *   x IS NOT NULL AND
+     *   length(arr) - countEqual(arr, x) - countEqual(arr, NULL) > 0
+     */
+    if (saop->useOr && optype == 2) {
```
theory, line 740: Ugh. Would be nice in a future patch to teach `chfdw_is_equal_op()` to use an enum.

src/deparse.c
```diff
@@ -504,6 +552,342 @@ agg_partial_kind(Aggref* agg) {
...
+        if (nulls == SAOP_ARR_NULL_FREE) {
+            return !exact_ctx || expr_never_null(probe, glob_cxt);
+        }
+        return !exact_ctx && expr_never_null(probe, glob_cxt);
+    }
+    return nulls == SAOP_ARR_NULL_FREE && expr_never_null(probe, glob_cxt);
+}
+
+/*
+ * How a NOT (x IN (SELECT ..)) — Postgres's NOT IN — can be shipped.
```
```suggestion
 * How to ship a Postgres `NOT (x IN (SELECT ..))`.
```

src/deparse.c
```diff
@@ -1215,17 +1642,42 @@ is_shippable_subplan(SubPlan* subplan, foreign_glob_cxt* glob_cxt) {
...
+     * diverges from Postgres only by yielding FALSE where Postgres computes
+     * NULL (a NULL probe stays NULL on both systems). That is invisible in a
+     * truth context but observable under NOT — which is how x NOT IN
+     * (SELECT ..) arrives here — and in value positions. Under one NOT in a
+     * truth context (EXPR_CTX_NEGATED) the guarded deparse form keeps
+     * Postgres's answer even when a NULL can reach the comparison, so only
+     * grouped/aggregated bodies fall back; a genuine value position demands
+     * exactness, which requires that no NULL can arrive from either side.
+     * See saop_null_semantics_ok's block comment for the divergence rules
+     * and deparseGuardedNotIn for the guarded form.
```
theory, line 1672:
```suggestion
     * ClickHouse evaluates IN under two-valued logic, so x IN (SELECT ..)
     * diverges from Postgres by yielding FALSE where Postgres computes
     * NULL (a NULL probe stays NULL on both systems). This behavior is invisible in a
     * truth context but observable under NOT --- which is how x NOT IN
     * (SELECT ..) arrives here --- and in value positions. Under one NOT in a
     * truth context (EXPR_CTX_NEGATED) the guarded deparse form keeps the
     * Postgres answer even when a NULL can reach the comparison, so only
     * grouped/aggregated bodies fall back; a genuine value position demands
     * exactness, which requires that no NULL can arrive from either side.
     * See saop_null_semantics_ok's block comment for the divergence rules
     * and deparseGuardedNotIn for the guarded form.
```

src/deparse.c
```diff
@@ -3318,6 +3800,60 @@ deparseSubPlanQuery(SubPlan* subplan, deparse_expr_cxt* context) {
...
+ *
+ *   CASE WHEN probe IS NULL
+ *        THEN NOT EXISTS (SELECT ..)                            -- empty?
+ *        ELSE probe NOT IN (SELECT ..)
+ *             AND NOT EXISTS (SELECT 1 .. AND (output IS NULL)) -- poison?
+ *   END
+ *
+ * Each piece a plan-time proof makes unnecessary is left out: a non-NULL
+ * probe drops the CASE (x NOT IN (empty) is already TRUE on both systems
+ * for a non-NULL probe), a non-NULL output column drops the poison guard.
```
theory: I think use of backticks around code bits would make this easier to read; I can't tell where the code ends, as the parens are all goofy.

test/expected/result_map.txt
```diff
@@ -187,6 +174,17 @@ import_schema.sql
...
+---------------------
+
+ Postgres | File
+----------|------------------------
+ 13-19    | in_null_semantics.out
+
+ ClickHouse | File
+------------|------------------------
+ 23+        | in_null_semantics.out
+
```
theory: Don't need this if one file covers all.

CHANGELOG.md
```diff
@@ -69,11 +77,20 @@ All notable changes to this project will be documented in this file. It uses the
...
+    `= ANY`/`<> ALL` over non-constant arrays matched NULLs as values.
+    These now ship only where PostgreSQL's three-valued answer is provably
+    preserved ([#313]).
+*   Fixed wrong results from pushed-down `<> ANY(array)`, which deparsed to
+    ClickHouse SQL computing `<> ALL` (`1 <> ANY('{1,5}')` is true — one
+    element differs — but returned false). The shape is now evaluated
+    locally ([#313]).
+*   Fixed a shippability walker bug that pushed down a `CASE arg WHEN ..`
+    expression without checking its branches when the tested expression was
+    itself unshippable ([#313]).
```
```suggestion
*   Fixed wrong results from pushed-down `= ANY`/`<> ALL` expressions and `IN`
    expressions over constant lists and when a `NULL` could reach the
    comparison, because ClickHouse evaluates `IN` under two-valued logic while
    PostgreSQL evaluates three-value `NULL` logic. Thees cases now ship only
    where PostgreSQL's three-valued answer can be provably preserved ([#313]).
*   Fixed wrong results from pushed-down `<> ANY(array)`, which deparsed to
    ClickHouse SQL computing `<> ALL` (`1 <> ANY('{1,5}')` is true — one
    element differs — but returned false). This shape evaluates locally
    ([#313]).
*   Fixed the pushdown of a `CASE arg WHEN ..` expression without checking its
    branches when the tested expression was itself unshippable ([#313]).
```

test/expected/in_null_semantics.out
```diff
@@ -0,0 +1,407 @@
+-- NULL semantics of the IN family (ScalarArrayOpExpr pushdown).
+--
+-- ClickHouse evaluates IN under two-valued logic: a probe that finds no
+-- match yields 0 even when the searched set contains a NULL, where Postgres
+-- computes NULL, and its has()/countEqual() functions match a NULL probe as
```
theory:
```suggestion
-- computes NULL; The ClickHouse has()/countEqual() functions also match a NULL probe as
```

test/expected/in_null_semantics.out
```diff
@@ -0,0 +1,407 @@
...
+-- if it were an ordinary value. A FALSE-for-NULL substitution is invisible
+-- while the result only qualifies rows (WHERE/JOIN/HAVING through AND/OR),
+-- but observable under NOT and anywhere the boolean's value is consumed.
+-- These tests pin which shapes deparse.c ships and that the rows returned
+-- match local (Postgres) semantics either way.
```
theory:
```suggestion
-- These tests ensure the shapes deparse.c ships and that the rows returned
-- match local (Postgres) semantics either way.
```

test/expected/in_null_semantics.out
```diff
@@ -0,0 +1,407 @@
...
+     1
+(1 row)
+
+-- ============================================================
+-- 5. Shippability-walker regressions fixed alongside the rules
+-- ============================================================
+-- A NULL array constant survives const-folding when the probe is a column;
+-- inspecting it for the IN-list deparse would detoast a null datum. Local.
+EXPLAIN (VERBOSE, COSTS OFF)
+SELECT id FROM tnull WHERE xn = ANY(NULL::int[]) ORDER BY id;
```
theory, line 360: Surely this always evaluates to false and returns no rows. Maybe we could avoid sending it to clickHouse at all.

test/expected/in_null_semantics.out
```diff
@@ -0,0 +1,407 @@
...
+EXPLAIN (VERBOSE, COSTS OFF)
+SELECT id FROM tnull WHERE CASE pg_backend_pid() WHEN 0 THEN xp = 1 ELSE xp > 0 END
+ORDER BY id;
+                                     QUERY PLAN                                     
+------------------------------------------------------------------------------------
+ Foreign Scan on in_null_test.tnull
+   Output: id
+   Filter: CASE pg_backend_pid() WHEN 0 THEN (tnull.xp = 1) ELSE (tnull.xp > 0) END
+   Remote SQL: SELECT id, xp FROM in_null_test.tnull ORDER BY id ASC NULLS LAST
+(4 rows)
```
theory, line 385: LOL, maybe we should have `pg_backend_pid()` push down a constant value. Useful for this test tho.

theory (approved):
If I read this right, we depend on the Postgres catalog to determine whether a var is `NULL`able. But it of course can lie: one can have a column marked `NOT NULL` in Postgres that actually has `NULL`s in ClickHouse. I think that's acceptable, given that users can update their foreign tables as appropriate.

doc/pg_clickhouse.md
```diff
@@ -1325,8 +1329,14 @@ row, and a NULL probe passes only against an empty set. Each guard is
 omitted when a `NOT NULL` declaration proves it unnecessary. The remaining
 shapes (NULL-capable `IN` expressions in select lists, `GROUP BY`, or
 `ORDER BY`, and `NOT IN` over arrays or grouped subqueries) are evaluated
-locally. Declaring columns `NOT NULL` maximizes pushdown; `IMPORT FOREIGN
-SCHEMA` does so automatically for non-`Nullable` ClickHouse columns.
+locally. Declaring columns `NOT NULL` maximizes pushdown; [IMPORT FOREIGN
+SCHEMA] does so automatically for non-`Nullable` ClickHouse columns.
+
+These rules assume ClickHouse's default `transform_null_in = 0`, which
+pg_clickhouse pins through the default value of the
```
theory: "pins" is the wrong word, since the user can change it.

src/option.c
```diff
@@ -688,7 +688,7 @@ _PG_init(void) {
         "Sets the default ClickHouse session settings.",
         NULL,
         &ch_session_settings,
-        "join_use_nulls 1, group_by_use_nulls 1, final 1",
+        "join_use_nulls 1, group_by_use_nulls 1, final 1, transform_null_in 0",
```
theory, line 689: Please mention this change in the changelog.

doc/pg_clickhouse.md
```diff
@@ -1310,6 +1314,30 @@ equivalents as follows:
...
+shapes (NULL-capable `IN` expressions in select lists, `GROUP BY`, or
+`ORDER BY`, and `NOT IN` over arrays or grouped subqueries) are evaluated
+locally. Declaring columns `NOT NULL` maximizes pushdown; [IMPORT FOREIGN
+SCHEMA] does so automatically for non-`Nullable` ClickHouse columns.
+
+These rules assume ClickHouse's default `transform_null_in = 0`, which
+pg_clickhouse sets on every query through the default value of the
+[`pg_clickhouse.session_settings`](#pg_clickhousesession_settings) parameter
+so that a ClickHouse server profile cannot silently change it. Overriding it
+there changes the semantics of every pushed `IN`.
```
```suggestion
so that a ClickHouse server profile cannot silently change it. Setting
`transform_null_in = 1` breaks the semantics of every pushed `IN`.
```

src/option.c
JoshDreamland: Mentioned.

CHANGELOG.md
```diff
@@ -14,6 +14,18 @@ All notable changes to this project will be documented in this file. It uses the
 *   Added the `clickhouse_server_version(server)` function, which reports the
     ClickHouse server version ("major.minor.patch") for a foreign server
     ([#293]).
+*   Added pushdown for subqueries the planner cannot flatten into joins
+    (SubPlans): correlated and uncorrelated scalar aggregate subqueries,
+    `EXISTS`, and single-column equality `IN`/`NOT IN`, in WHERE and HAVING.
```
```suggestion
    `EXISTS`, and single-column equality `IN`/`NOT IN`, in `WHERE` and `HAVING`.
```

src/deparse.c
JoshDreamland: I'd use it after PG 16 support drops. Otherwise it's two (relatively small) codepaths doing the same thing.

CHANGELOG.md
```diff
@@ -74,6 +86,17 @@ All notable changes to this project will be documented in this file. It uses the
...
     and unbounded `DateTime64` scale lookups ([#313]).
+*   Fixed wrong results from pushed-down `= ANY`/`<> ALL` expressions and `IN`
+    expressions over constant lists and when a `NULL` could reach the
+    comparison, because ClickHouse evaluates `IN` under two-valued logic while
+    PostgreSQL evaluates three-value `NULL` logic. These cases now ship only
+    where PostgreSQL's three-valued answer can be provably preserved ([#315]).
+*   Fixed wrong results from pushed-down `<> ANY(array)`, which deparsed to
+    ClickHouse SQL computing `<> ALL` (`1 <> ANY('{1,5}')` is true — one
+    element differs — but returned false). This shape evaluates locally
+    ([#315]).
```
theory: I still don't understand what this is saying, and the mdash signals text extrusion rather than explanation.

theory (approved):
Add yourself also co-author to the README!

src/deparse.c
JoshDreamland: Added a comment. Let me know if I should elaborate in the comment (or just send me a patch), but basically, the idea is that we give proving nullability an honest try (where "honest try" here means "the trivial case"), but it's okay if we're wrong, because if we can't prove that the column can never be null, we just emit our uglier/more verbose query that ensures the behavior is the same regardless. But if we can prove nullability, we emit the prettier, sleeker, and probably slightly more performant (though I cba to benchmark rn) flavor.

src/deparse.c
JoshDreamland: Agreed. This is already on my follow-up list. I won't drop it unless my extreme ADHD pairs with a breakage in Claude's task list (I know you aren't the biggest fan of LLMs, but they're paradise for my specific flavor of scatter-brainedness).

test/expected/in_null_semantics.out
JoshDreamland: PG doesn't fold it, and I didn't spend code folding it because who tf would write this, but this PR goes as far as to remove the crash (by not shipping). Sound good?

I.e. yes, it always returns false, yes, we could optimize it, no, PG doesn't and I haven't, because who cares? Bug me for a follow-up if the answer is "you."

test/expected/in_null_semantics.out
JoshDreamland: Haha. Don't go defeating my fixtures, now. But if you have a better unshippable-yet-stable function, I'm all ears.

src/deparse.c
JoshDreamland: Yeah, I think you've discovered this in your own exploration, but for posterity: even `join_use_nulls` isn't the full solution here. The `transform_null_in` option is the closest cousin to what we need, but it doesn't do the three-value logic we need to match Postgres.

src/deparse.c
JoshDreamland: Took your suggestion.

CHANGELOG.md
JoshDreamland: It does love those, yes; my earlier instructions were "use my text from the PR description," but I guess that bears repeating. I'll fold it in.

src/deparse.c
theory: Yah, hence my comment on ClickHouse/ClickHouse#34737.

src/deparse.c
theory: Hrm. I tend to prefer supporting both, that way I don't have to figure it out again when it's time to drop 16, just drop it.

src/deparse.c
theory: I prefer your GH comment to your C comment lol.

test/expected/in_null_semantics.out
theory: Heh, it's not, I argued against making the same sort of optimization when Philip suggested it. I just thought it might be less complicated here. Fine the way it is.

doc/pg_clickhouse.md
```diff
@@ -795,8 +795,12 @@ settings] to be set on subsequent queries. Example:
 SET pg_clickhouse.session_settings = 'join_use_nulls 1, final 1';
 ```
 
-The default is `join_use_nulls 1, group_by_use_nulls 1, final 1`. Set it to an
-empty string to fall back on the ClickHouse server's settings.
+The default is `join_use_nulls 1, group_by_use_nulls 1, final 1,
+transform_null_in 0`. Set it to an empty string to fall back on the
+ClickHouse server's settings — but note that pushdown correctness depends on
```
theory, line 800:
```suggestion
ClickHouse server's settings, but note that pushdown correctness depends on
```

doc/pg_clickhouse.md
```diff
@@ -1310,6 +1314,30 @@ equivalents as follows:
...
+shapes (NULL-capable `IN` expressions in select lists, `GROUP BY`, or
+`ORDER BY`, and `NOT IN` over arrays or grouped subqueries) are evaluated
+locally. Declaring columns `NOT NULL` maximizes pushdown; [IMPORT FOREIGN
+SCHEMA] does so automatically for non-`Nullable` ClickHouse columns.
+
+These rules assume ClickHouse's default `transform_null_in = 0`, which
+pg_clickhouse sets on every query through the default value of the
+[`pg_clickhouse.session_settings`](#pg_clickhousesession_settings) parameter
+so that a ClickHouse server profile cannot silently change it. Setting
+`transform_null_in = 1` breaks the semantics of every pushed `IN`.
```
theory, line 1339:
```suggestion
These rules assume ClickHouse's default `transform_null_in = 0`, which
pg_clickhouse sets on every query through the
[`pg_clickhouse.session_settings`](#pg_clickhousesession_settings) default
so that a ClickHouse server profile cannot change it. Setting
`transform_null_in = 1` breaks the semantics of every pushed `IN`.
```

PR #316, opened 2026-07-16
Support inserting Array(Nullable(T)) via binary driver

serprex:
requires https://github.com/ClickHouse/clickhouse-c/pull/18

### DIFF
--- a/src/binary/insert.c
+++ b/src/binary/insert.c
@@ -1,13 +1,13 @@
 /*
  * insert.c
  *
- * INSERT path for the binary driver. begin_insert sends the query, takes
- * the server's empty Data block as schema, classifies each column into
- * one of a handful of accumulator layouts (fixed / string / LC / array
- * variants), and exposes typed ch_binary_append_* shims. flush_block
- * materialises the accumulators into a chc_block_builder and ships one
- * Data packet; finalize_insert sends the closing empty Data and drains.
- * release_insert is reset-callback safe: it never talks to the wire.
+ * Handle binary INSERT lifecycle
+ *
+ *  -begin_insert sends query and uses server's empty Data block as schema
+ * - append functions buffer column values in format chosen from schema
+ * - flush_block assembles buffered columns and sends one Data packet
+ * - finalize_insert sends closing empty Data packet and reads final server response
+ * - release_insert can run during memory context reset and never writes to connection.
  */
 
 #include "postgres.h"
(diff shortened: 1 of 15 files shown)


src/binary/insert.c
```diff
@@ -1,13 +1,13 @@
...
- * materialises the accumulators into a chc_block_builder and ships one
- * Data packet; finalize_insert sends the closing empty Data and drains.
- * release_insert is reset-callback safe: it never talks to the wire.
+ * Handle binary INSERT lifecycle
+ *
+ * begin_insert sends query and uses server's empty Data block as schema
+ * Append functions buffer column values in format chosen from schema
+ * flush_block assembles buffered columns and sends one Data packet
+ * finalize_insert sends closing empty Data packet and reads final server response
+ * release_insert can run during memory context reset and never writes to connection.
```
theory:
```suggestion
 *  - begin_insert sends query and uses server's empty Data block as schema
 *  - append functions buffer column values in format chosen from schema
 *  - flush_block assembles buffered columns and sends one Data packet
 *  - finalize_insert sends closing empty Data packet and reads final server response
 *  - release_insert can run during memory context reset and never writes to connection.
```

src/binary/insert.c
```diff
@@ -1246,6 +1239,87 @@ build_lc_dict(
     *out_n_rows    = n_rows;
 }
 
+static chc_column*
```
theory: inline?

src/binary/insert.c
```diff
@@ -1246,6 +1239,87 @@ build_lc_dict(
...
+    case IC_STRING: {
+        size_t n_rows = c->body_offs.len;
+        chc_column* leaf =
+            col_node(chc_build_string(c->body_offs.data, c->body.data, n_rows));
+
+        return c->is_nullable ? col_node(chc_build_nullable(c->nulls.data, leaf))
+                              : leaf;
+    }
+    case IC_JSON_STRING:
+        /* clickhouse-c adds 8-byte JSON prefix, then writes String-Binary rows */
```
theory: The implication being that it never builds a nullable?

src/binary/insert.c
```diff
@@ -1246,6 +1239,87 @@ build_lc_dict(
...
+        /* Build nested arrays inside out, use outermost offsets last */
+        for (int lvl = c->ndim - 1; lvl >= 0; lvl--) {
+            u64buf* offs = lvl == 0 ? &c->arr_offs : &c->arr_offs_inner[lvl - 1];
+
+            child = col_node(chc_build_array(offs->data, offs->len, child));
+        }
+        return child;
+    }
+    }
+    pg_unreachable();
```
theory, line 1333: Ooh, I wasn't aware of this function.

theory (approved):
Does seem to simplify things.

src/binary/insert.c
serprex: `Nullablle(Object('json'))` JSON isn't allowed, but `Nullable(JSON)` with modern type works. Fixed & added test `binary_nullable_json.sql`. Ended up revising ic_layout to mirror clickhouse-c chc_column nodes

PR #317, opened 2026-07-17
Push down the array IN family unconditionally

JoshDreamland:
The array IN family (`IN`, `NOT IN`, `= ANY`, `= ALL`, `<> ANY`, `<> ALL`) now ships unconditionally, in every context, instead of falling back to local evaluation whenever non-nullability couldn't be proven.

`saop_null_semantics_ok` collapses to true (past the pre-existing empty-array and NULL-array-constant checks); deparseScalarArrayOpExpr picks the cheap native/`has()`/`countEqual()` form when it can prove neither side is NULL, and a unified `deparseGuardedScalarArrayOp` helper otherwise; see doc comment thereabove for details about guard mechanics.

The non-nullability proof also now follows basic arithmetic (`+`, `-`, `*`, unary `-`) over non-NULL constants and NOT NULL columns, via a hand-picked list of operator functions: being strict alone doesn't prove a function can't return NULL from non-NULL inputs (->> on a missing JSON key is a counterexample), so each entry is individually
checked. The list is sorted once at load time (a constructor-attribute function, since these OIDs aren't guaranteed identical across every supported PG version) so the check is a bsearch instead of a linear scan.

A trade-off (mitigated in #321): a constant list with a possible NULL previously shipped via native `IN(...)` specifically in a truth context, where FALSE-for-NULL substitution from ClickHouse makes no difference. The deparse function has no way to see context as of this commit, so it now always guards that case instead, which sucks, but buys us pushdown for the rest of the array IN family.

Also, finally patches `chfdw_is_equal_op` to return a `ChfdwEqualOp` enum (NONE/EQ/NE) instead of magic 0/1/2 ints.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@ All notable changes to this project will be documented in this file. It uses the
     `pg_clickhouse.session_settings` parameter, so that a ClickHouse server
     profile cannot silently change the `IN` semantics the pushdown rules rely
     on ([#315]).
+*   Added or corrected pushdown for `IN`-family operations on arrays. Previous
+    behavior was missing or incorrect, either structurally or due to behavioral
+    differences in the ClickHouse implementation of the operation. Corrected
+    the structurally incorrect pushdown ([#315]) and handled ClickHouse
+    behavioral changes using special case statements where needed ([#317]).
 *   Added pushdown for more aggregate functions ([#290]):
     *   [corr]
     *   [covarpop]
@@ -92,8 +97,7 @@ All notable changes to this project will be documented in this file. It uses the
 *   Fixed wrong results from pushed-down `= ANY`/`<> ALL` expressions and `IN`
     expressions over constant lists and when a `NULL` could reach the
     comparison, because ClickHouse evaluates `IN` under two-valued logic while
-    PostgreSQL evaluates three-value `NULL` logic. These cases now ship only
-    where PostgreSQL's three-valued answer can be provably preserved ([#315]).
+    PostgreSQL evaluates three-value `NULL` logic ([#315], [#317]).
 *   Fixed `<> ANY(array)` deparsing to ClickHouse SQL that computes `<> ALL`,
     which is wrong even with no `NULL`s involved: In Postgres,
     `1 <> ANY('{1,5}')` is `TRUE`. For now, do not push down ([#315]).
@@ -149,6 +153,8 @@ All notable changes to this project will be documented in this file. It uses the
     "ClickHouse/pg_clickhouse#316 Support inserting Array(Nullable(T)) via binary driver"
   [#318]: https://github.com/ClickHouse/pg_clickhouse/pull/318
     "ClickHouse/pg_clickhouse#318 Push down the re2 v0.4 `@~` operator"
+  [#317]: https://github.com/ClickHouse/pg_clickhouse/pull/317
+    "ClickHouse/pg_clickhouse#317 Push down the array IN family unconditionally"
 
 ## [v0.3.2] — 2026-06-16
 
--- a/src/deparse.c
+++ b/src/deparse.c
@@ -174,7 +174,7 @@ typedef struct deparse_expr_cxt {
 
 /*
  * How an expression's result is consumed, for the IN-family NULL-semantics
- * rules (see the block comment for saop_null_semantics_ok).
+ * rules (see the block comment for deparseScalarArrayOpExpr).
  *
  * EXPR_CTX_TRUTH    only qualifies rows: NULL and FALSE are interchangeable.
  * EXPR_CTX_NEGATED  the direct operand of one NOT whose own result is in a
@@ -185,6 +185,12 @@ typedef struct deparse_expr_cxt {
  *                   exactly which SubPlans get the guarded form.
  * EXPR_CTX_EXACT    the value itself is observable: exact three-valued
  *                   equivalence required.
+ *
+ * NOTE: deparseExpr's own switch on nodeTag degrades its analogous
+ * deparse_expr_cxt.truth_ctx to false before dispatch for every node type
+ * except T_BoolExpr/T_CaseExpr/T_RelabelType/T_ScalarArrayOpExpr -- the same
+ * types this walker does something other than force EXPR_CTX_EXACT for. If
+ * you add a type here that preserves ctx for its children, add it there too.
  */
 typedef enum ExprTruthCtx {
     EXPR_CTX_TRUTH,
@@ -459,12 +465,11 @@ chfdw_is_foreign_expr(
     return true;
 }
 
-/* 1: '=', 2: '<>', 0 - false */
-int
+CHEqualOp
 chfdw_is_equal_op(Oid opno) {
     Form_pg_operator operform;
     HeapTuple opertup;
-    int res = 0;
+    CHEqualOp res = CH_OP_NONE;
 
     opertup = SearchSysCache1(OPEROID, ObjectIdGetDatum(opno));
     if (!HeapTupleIsValid(opertup)) {
@@ -474,11 +479,11 @@ chfdw_is_equal_op(Oid opno) {
     operform = (Form_pg_operator)GETSTRUCT(opertup);
 
     if (NameStr(operform->oprname)[0] == '=' && NameStr(operform->oprname)[1] == '\0') {
-        res = 1;
+        res = CH_OP_EQ;
     } else if (
         NameStr(operform->oprname)[0] == '<' && NameStr(operform->oprname)[1] == '>'
     ) {
-        res = 2;
+        res = CH_OP_NE;
     }
 
     ReleaseSysCache(opertup);
@@ -555,10 +560,14 @@ agg_partial_kind(Aggref* agg) {
 static bool
 is_ok_in_expr(Expr* expr);
 
+static bool
+op_expr_never_null(OpExpr* op, foreign_glob_cxt* glob_cxt);
+
 /*
- * True if `expr` provably cannot evaluate to NULL: a non-NULL constant, or a
+ * True if `expr` provably cannot evaluate to NULL: a non-NULL constant, a
  * column of a scanned relation that is declared NOT NULL and cannot have been
- * null-extended by an outer join. Anything else is assumed nullable.
+ * null-extended by an outer join, or certain arithmetic operators over such
+ * inputs (see op_expr_never_null). Anything else is assumed nullable.
  *
  * PG 17+ precomputes the catalog half of this as
  * RelOptInfo->notnullattnums; the syscache lookup below keeps one code path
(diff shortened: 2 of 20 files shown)


src/include/fdw.h
```diff
@@ -358,11 +358,19 @@ chfdw_get_jointype_name(JoinType jointype);
 char*
 chfdw_array_to_ch_literal(Datum arr);
 
+/* chfdw_is_equal_op's classification of a comparison operator's name. */
+typedef enum ChfdwEqualOp {
```
theory: Should be `CHEqualOp` to match other enums like `CHRemoteTableEngine`.

CHANGELOG.md
```diff
@@ -26,6 +26,19 @@ All notable changes to this project will be documented in this file. It uses the
     `pg_clickhouse.session_settings` parameter, so that a ClickHouse server
     profile cannot silently change the `IN` semantics the pushdown rules rely
     on ([#315]).
+*   Corrected pushdown for `<> ANY(array)`. Ships an element-counting form
+    rather than `not has()`, which actually computes `<> ALL`. This was briefly
+    corrected by suspending pushdown ([#315]) but then fixed properly ([#317]).
```
theory: Since we didn't release the "briefly corrected" bit, omit any reference to that state. It never shipped.

CHANGELOG.md
```diff
@@ -26,6 +26,19 @@ All notable changes to this project will be documented in this file. It uses the
...
+    Like with `NOT IN`, a more elegant spelling is used when non-nullability can
+    be proven, and in other cases, an uglier shape is used which can handle NULL
+    values correctly (matching Postgres behavior).
+*   Extended that guarded shape to the rest of the array `IN` family
+    (`IN`/`NOT IN`, `= ANY`, `= ALL`, `<> ALL`), so all of them now push down
+    unconditionally, in every context (`SELECT` list and `GROUP BY` included),
+    instead of falling back to local evaluation whenever non-nullability
+    couldn't be proven. The non-nullability proof that selects the cheaper
+    unguarded form now also follows basic arithmetic (`+`, `-`, `*`, unary
+    `-`) over non-NULL constants and `NOT NULL` columns ([#317]).
```
theory: This might all be more detail than necessary TBH. Just say what changed and the impact.

Though I'll likely edit it all again during release prep.

test/expected/in_null_semantics.out
```diff
@@ -312,20 +433,19 @@ SELECT xn IN (1, 500) AS flag, count(*) FROM tnull GROUP BY flag ORDER BY flag;
...
-   ->  HashAggregate
-         Output: ((xn = ANY ('{1,NULL}'::integer[]))), count(*)
-         Group Key: (tnull.xn = ANY ('{1,NULL}'::integer[]))
-         ->  Foreign Scan on in_null_test.tnull
-               Output: (xn = ANY ('{1,NULL}'::integer[]))
-               Remote SQL: SELECT xn FROM in_null_test.tnull
-(9 rows)
+   Relations: Aggregate on (tnull)
+   Remote SQL: SELECT (CASE WHEN xn IS NULL AND notEmpty([1,NULL]) THEN NULL WHEN countEqual([1,NULL],xn) > 0 THEN true WHEN countEqual([1,NULL], NULL) > 0 THEN NULL ELSE false END), count(*) FROM in_null_test.tnull GROUP BY ((CASE WHEN xn IS NULL AND notEmpty([1,NULL]) THEN NULL WHEN countEqual([1,NULL],xn) > 0 THEN true WHEN countEqual([1,NULL], NULL) > 0 THEN NULL ELSE false END)) ORDER BY (CASE WHEN xn IS NULL AND notEmpty([1,NULL]) THEN NULL WHEN countEqual([1,NULL],xn) > 0 THEN true WHEN countEqual([1,NULL], NULL) > 0 THEN NULL ELSE false END) ASC NULLS LAST
+(4 rows)
```
theory, line 465: Oh this is nice!

src/deparse.c
```diff
@@ -474,11 +473,11 @@ chfdw_is_equal_op(Oid opno) {
     operform = (Form_pg_operator)GETSTRUCT(opertup);
 
     if (NameStr(operform->oprname)[0] == '=' && NameStr(operform->oprname)[1] == '\0') {
-        res = 1;
+        res = CHFDW_OP_EQ;
     } else if (
         NameStr(operform->oprname)[0] == '<' && NameStr(operform->oprname)[1] == '>'
     ) {
-        res = 2;
+        res = CHFDW_OP_NE;
```
theory: I'm sure the parser makes it the same, but add a test to ensure that `!=` gets to us as `<>`.

src/deparse.c
```diff
@@ -1071,14 +1068,25 @@ foreign_expr_walker(Node* node, foreign_glob_cxt* glob_cxt, ExprTruthCtx ctx) {
...
+        arr = (Expr*)lsecond(oe->args);
+
+        /*
+         * A NULL array constant survives const-folding when the probe is
+         * not constant, and the operator's strictness makes the result
+         * NULL for every row. Keep it local: no deparse form models that,
+         * and deparseScalarArrayOpExpr's is_ok_in_expr call cannot inspect
+         * a NULL array datum. Otherwise this always ships; see
+         * deparseScalarArrayOpExpr's block comment for how it preserves
+         * Postgres's NULL semantics either way.
```
theory: I can't parse this. In the first sentence, is the "and" clause part of the "when" expression or the result of its evaluation? The second and third sentences are no better.

src/deparse.c
```diff
@@ -1071,14 +1068,25 @@ foreign_expr_walker(Node* node, foreign_glob_cxt* glob_cxt, ExprTruthCtx ctx) {
...
+        Expr* arr;
 
-        if (!optype) {
+        if (optype == CHFDW_OP_NONE) {
             return false;
         }
 
-        /* IN under two-valued ClickHouse logic; see saop_null_semantics_ok */
-        if (!saop_null_semantics_ok(oe, optype, ctx, glob_cxt)) {
+        arr = (Expr*)lsecond(oe->args);
```
theory: Move below following comment.

src/deparse.c
```diff
@@ -5199,19 +5207,156 @@ is_ok_in_expr(Expr* expr) {
...
+    deparseExpr(expr, context);
+    context->buf = buf;
+    return capture.data;
+}
 
-    if (optype == 0) {
+/*
+ * Emit a guarded CASE reproducing Postgres's exact three-valued answer for
+ * any (useOr, optype) combination of the IN family, checking at runtime
+ * rather than requiring a plan-time proof:
```
theory, line 5242:
```suggestion
 * Emit a guarded CASE reproducing Postgres's exact three-valued answer for
 * any combination of the IN family (useOr, optype), checking at runtime
 * rather than requiring a plan-time proof:
```

src/deparse.c
```diff
@@ -5199,19 +5207,156 @@ is_ok_in_expr(Expr* expr) {
...
+ * `notEmpty()` check in the first branch, so the ELSE answers.
+ *
+ * A NULL probe against a non-empty array is NULL on both systems.
+ *
+ * The actual predicate implementation for the form above must use `countEqual`,
+ * and there is no `countNotEqual`, which means that complemented forms must
+ * instead compute their count by subtracting from the full array length.
+ * Here, Postgres's `useOr` determines most of our logic, while optype logically
+ * determines which operation we apply to each element but practically instead
+ * determines how we use the result of `countEqual()`.
```
theory: I don't understand the meaning starting with "but practically".

src/deparse.c
```diff
@@ -5199,19 +5207,156 @@ is_ok_in_expr(Expr* expr) {
...
+            "(length(%s) - countEqual(%s,%s) - countEqual(%s, NULL)) > 0",
+            arrsql,
+            arrsql,
+            probesql,
+            arrsql
+        );
+    }
+    appendStringInfo(
+        buf,
+        " THEN %s WHEN countEqual(%s, NULL) > 0 THEN NULL ELSE %s END)",
```
theory, line 5304: I think we should be consistent in the use of space after comma. I believe it' usually there, as in this line, so ought to be added to the above two `countEqual` lines.

src/deparse.c
```diff
@@ -5199,19 +5207,156 @@ is_ok_in_expr(Expr* expr) {
...
+ *
+ * Postgres/Standard SQL usually disagrees:
+ *   1    in {1, 2, 3}:   1
+ *   10   in {1, 2, 3}:   0
+ *   1    in {1, NULL}:   1
+ *   10   in {1, NULL}:   NULL  <-- The problem case
+ *   NULL in {1, NULL}:   NULL
+ *   NULL in {1, 2, 3}:   NULL
+ *
+ * The two systems agree when NULL is not involved, but in SQL, NULL serves as
```
```suggestion
 * The two systems agree when NULL is not involved, but in standard SQL, NULL serves as
```

src/deparse.c
```diff
@@ -5199,19 +5207,156 @@ is_ok_in_expr(Expr* expr) {
...
+ * the divergent cases.
+ *
+ * ClickHouse has a separate behavioral quirk for arrays, similar to its `IN`
+ * mechanics: `has()` and `countEqual()` will match NULL strictly against NULL
+ * in those contexts as well, and are unaffected by `transform_null_in` or
+ * similar settings. This means these routines cannot directly be used to
+ * implement `<> ALL()` / `= ANY()`, but it also gives us a correct (if uglier)
+ * pushdown spelling.
+ *
+ * We force `transform_null_in=0` to cover the cases it can, then in cases where
```
theory: I'm thinking we should do a bit of refactoring in a future commit to always set `transform_null_in=0` and any other settings that we depend on. The http driver already does this for formatting settings its TSV parser requires, and the binary driver does it for `output_format_native_write_json_as_string=1`.

```suggestion
 * We force `transform_null_in=0` to cover the cases it can, and in cases where
```

src/deparse.c
```diff
@@ -5199,19 +5207,156 @@ is_ok_in_expr(Expr* expr) {
...
+ * values are nullable (or we cannot prove that they aren't), emit extra guard
```
```suggestion
 * values are nullable (or we cannot prove that they aren't), emit the extra guard
```

src/deparse.c
```diff
@@ -5199,19 +5207,156 @@ is_ok_in_expr(Expr* expr) {
...
+ * branches below to ensure Postgres behavior is preserved after pushdown.
+ *
+ * Deparse given ScalarArrayOpExpr expression. Every branch below
```
theory: Seems like this first sentence should be the first sentence of the whole comment block, no?

src/deparse.c
```suggestion
                "pg_clickhouse supports only equal and not equal operations on ANY/ALL "
```

src/deparse.c
```diff
@@ -5222,66 +5367,67 @@ deparseScalarArrayOpExpr(ScalarArrayOpExpr* node, deparse_expr_cxt* context) {
...
+     * otherwise both the array and the probe must be provably non-NULL.
+     */
+    cheap_ok = nulls == SAOP_ARR_EMPTY ||
+               (nulls == SAOP_ARR_NULL_FREE && expr_never_null(arg1, &glob_cxt));
 
-            /* Deparse left operand. */
-            arg1 = linitial(node->args);
-            deparseExpr(arg1, context);
+    /*
+     * very narrow case for IN()/NOT IN(): native form is exact once the
```
```suggestion
     * Very narrow case for IN()/NOT IN(): the native form is exact once the
```

src/deparse.c
```diff
@@ -5222,66 +5367,67 @@ deparseScalarArrayOpExpr(ScalarArrayOpExpr* node, deparse_expr_cxt* context) {
...
-            /* Deparse right operand. */
-            deparseExpr(arg2, context);
-            appendStringInfoChar(buf, ',');
+    /*
+     * An empty array is exact via the cheap form below regardless of the
+     * probe's nullability (has()/countEqual() degenerate correctly on an
+     * empty array by arithmetic), so it counts as "cheap ok" on its own;
+     * otherwise both the array and the probe must be provably non-NULL.
+     */
+    cheap_ok = nulls == SAOP_ARR_EMPTY ||
```
theory: Move to calculate after the "very narrow case" below.

theory (approved):
This looks great, mostly just nits and desiring a bit more clarity of narrative text.

src/deparse.c
JoshDreamland: Rewrote; better?

src/deparse.c
JoshDreamland: How about "in practice"?

src/deparse.c
JoshDreamland: Yeah, good catch; looks like a copy/paste failure. I had this block on the deleted routine originally.

src/include/fdw.h
JoshDreamland: Renamed.

src/deparse.c
JoshDreamland: Correct; added.

src/deparse.c
JoshDreamland: Moved.

CHANGELOG.md
JoshDreamland: Sounds good; done.

CHANGELOG.md
JoshDreamland: Simplified. How's that?

src/deparse.c
JoshDreamland: Fixed, I think (there were a lot; may have missed some).

src/deparse.c
JoshDreamland: Moved.

CHANGELOG.md
theory: Much better; the `IN` change should be roughly the same, yes?

src/deparse.c
theory: Thank you, yes, although `x in NULL::int[]` isn't real syntax. Do you mean `x = ANY( NULL::int[])` or `x IN (NULL)`?

src/deparse.c
theory: Much clearer, thanks.

src/deparse.c
theory: `/,\S/` doesn't turn up any, so good!

PR #319, opened 2026-07-21
Fix foreign table selection for pushed-down joins

k-bx:
## Problem

For pushed-down join or upper-relation scans, `scanrelid` is zero.
`clickhouseBeginForeignScan()` selected the lowest member of `fs_relids` and
assumed it represented a base foreign table.

With pulled-up subqueries and an outer join, that member can instead be a
synthetic join RTE whose relation OID is invalid. Execution then calls
`GetForeignTable(InvalidOid)` and fails with:

`cache lookup failed for foreign table 0`

Plain `EXPLAIN` does not expose the problem because foreign-scan initialization
is skipped for explain-only execution.

## Fix

- On PostgreSQL 16+, use the core-populated `ForeignScan.checkAsUser` and
  `ForeignScan.fs_server` fields directly, without looking up a representative
  range-table entry.
- On PostgreSQL 13–15, scan `fs_relids` until a foreign-table RTE is found so
  its `checkAsUser` value can be used.
- Use the core-populated `fs_server` field for the user mapping on every
  supported PostgreSQL version.
- Report an explicit error if no representative foreign table exists on
  PostgreSQL 13–15.

## Test

Adds a regression covering execution of a pushed-down aggregate over pulled-up
subqueries with mixed inner and left joins.

Validated with the project pre-commit checks, full PostgreSQL 14 and 18 builds,
and an end-to-end PostgreSQL 18/ClickHouse execution test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -103,6 +103,9 @@ All notable changes to this project will be documented in this file. It uses the
     `1 <> ANY('{1,5}')` is `TRUE`. For now, do not push down ([#315]).
 *   Fixed the pushdown of a `CASE arg WHEN ..` expression without checking its
     branches when the tested expression was itself unshippable ([#315]).
+*   Fixed an error when incorrectly selecting an invalid relation OID when
+    selecting the user to execute the remote query. Thanks to Kostia R for the
+    PR ([#319])!
 
 ### 📚 Documentation
 
@@ -155,6 +158,8 @@ All notable changes to this project will be documented in this file. It uses the
     "ClickHouse/pg_clickhouse#318 Push down the re2 v0.4 `@~` operator"
   [#317]: https://github.com/ClickHouse/pg_clickhouse/pull/317
     "ClickHouse/pg_clickhouse#317 Push down the array IN family unconditionally"
+  [#319]: https://github.com/ClickHouse/pg_clickhouse/pull/319
+    "ClickHouse/pg_clickhouse#319 Fix foreign scan RTE selection"
 
 ## [v0.3.2] — 2026-06-16
 
--- a/src/fdw.c
+++ b/src/fdw.c
@@ -1134,11 +1134,12 @@ clickhouseBeginForeignScan(ForeignScanState* node, int eflags) {
     ForeignScan* fsplan = (ForeignScan*)node->ss.ps.plan;
     EState* estate      = node->ss.ps.state;
     ChFdwScanState* fsstate;
+#if PG_VERSION_NUM < 160000
     RangeTblEntry* rte;
-    Oid userid;
-    ForeignTable* table;
-    UserMapping* user;
     int rtindex;
+#endif
+    Oid userid;
+    UserMapping* user;
     int numParams;
 
     /*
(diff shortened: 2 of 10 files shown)


src/fdw.c
```diff
@@ -1157,13 +1157,27 @@ clickhouseBeginForeignScan(ForeignScanState* node, int eflags) {
     /*
      * Identify which user to do the remote access as. This should match what
      * ExecCheckRTEPerms() does. In case of a join or aggregate, use the
-     * lowest-numbered member RTE as a representative; we would get the same
-     * result from any.
+     * foreign-table member RTE as a representative; we would get the same
+     * result from any. fs_relids can also contain synthetic join RTEs, which
+     * do not have a relation OID.
```
theory:
```suggestion
     * foreign-table member RTE as a representative (we would get the same
     * result from any). fs_relids can also contain synthetic join RTEs, which
     * do not have a relation OID.
```

theory (approved):
Nice find, thank you!

src/fdw.c
```diff
@@ -1157,13 +1157,27 @@ clickhouseBeginForeignScan(ForeignScanState* node, int eflags) {
...
      * ExecCheckRTEPerms() does. In case of a join or aggregate, use the
-     * lowest-numbered member RTE as a representative; we would get the same
-     * result from any.
+     * foreign-table member RTE as a representative. fs_relids can also
+     * contain synthetic join RTEs, which do not have a relation OID.
      */
     if (fsplan->scan.scanrelid > 0) {
         rtindex = fsplan->scan.scanrelid;
     } else {
         rtindex = bms_next_member(fsplan->fs_relids, -1);
```
serprex: odd having dead assignment to rtindex, is this call needed at all?

test/sql/binary_queries.sql
```diff
@@ -197,6 +197,17 @@ SELECT COUNT(DISTINCT c1) FILTER (WHERE c1 < 20) FROM ft2;
 EXPLAIN (VERBOSE, COSTS OFF) SELECT COUNT(*) FILTER (WHERE c1 < 20) FROM ft2;
 SELECT COUNT(*) FILTER (WHERE c1 < 10) FROM ft2;
 
+-- Pulled-up subqueries can place an outer-join RTE before the base foreign
+-- tables represented by a pushed-down scan.
+SELECT t4.c1, COUNT(DISTINCT t4.c2)
```
serprex, line 212: let's also `EXPLAIN (VERBOSE, COSTS OFF)`

CHANGELOG.md
```diff
@@ -99,6 +99,9 @@ All notable changes to this project will be documented in this file. It uses the
     `1 <> ANY('{1,5}')` is `TRUE`. For now, do not push down ([#315]).
 *   Fixed the pushdown of a `CASE arg WHEN ..` expression without checking its
     branches when the tested expression was itself unshippable ([#315]).
+*   Fixed an error when incorrectly selecting an invalid relation OID when
+    selecting the user to execute the remote query. Thanks to Kostia R for the
+    PR ([319])!
```
serprex:
```suggestion
    PR ([#319])!
```

src/fdw.c
```diff
@@ -1157,13 +1157,27 @@ clickhouseBeginForeignScan(ForeignScanState* node, int eflags) {
     /*
      * Identify which user to do the remote access as. This should match what
      * ExecCheckRTEPerms() does. In case of a join or aggregate, use the
-     * lowest-numbered member RTE as a representative; we would get the same
-     * result from any.
+     * foreign-table member RTE as a representative. fs_relids can also
+     * contain synthetic join RTEs, which do not have a relation OID.
      */
     if (fsplan->scan.scanrelid > 0) {
```
serprex, line 1168: this can be simpler, PG16 & above we can use `fsplan->checkAsUser`

src/fdw.c
theory: That's my fault, pasto from validating the test failed with the old code.

src/fdw.c
theory: Are you suggesting:

```patch
-    if (fsplan->scan.scanrelid > 0) {
-        rtindex = fsplan->scan.scanrelid;
+    if (fsplan->checkAsUser > 0) {
+        rtindex = fsplan->checkAsUser;
```

theory: LOL no, that doesn't work.

theory: This what currently compiles on Postgres 16+:

```c
    if (fsplan->scan.scanrelid > 0) {
        rtindex = fsplan->scan.scanrelid;
    } else {
        rtindex = bms_next_member(fsplan->fs_base_relids, -1);
        if (rtindex < 0) {
            elog(ERROR, "could not find foreign table for pg_clickhouse scan");
        }
    }
    rte    = rt_fetch(rtindex, estate->es_range_table);
    userid = OidIsValid(fsplan->checkAsUser) ? fsplan->checkAsUser : GetUserId();
```

serprex:
```c
 #if PG_VERSION_NUM >= 160000
      userid =
          OidIsValid(fsplan->checkAsUser) ? fsplan->checkAsUser : GetUserId();
  #else
      if (fsplan->scan.scanrelid > 0) {
          rtindex = fsplan->scan.scanrelid;
      } else {
          rtindex = -1;
          while ((rtindex = bms_next_member(fsplan->fs_relids, rtindex)) >= 0) {
              rte = rt_fetch(rtindex, estate->es_range_table);
              if (rte->rtekind == RTE_RELATION &&
                  rte->relkind == RELKIND_FOREIGN_TABLE) {
                  break;
              }
          }
          if (rtindex < 0) {
              elog(ERROR, "could not find foreign table for pg_clickhouse scan");
          }
      }

      rte = rt_fetch(rtindex, estate->es_range_table);
      userid = rte->checkAsUser ? rte->checkAsUser : GetUserId();
  #endif

      user = GetUserMapping(userid, fsplan->fs_server);
```

theory: You need to assign to `rte` on 16+. This does not.

serprex: you don't need rte/table on 16+

k-bx: Applied in `2d35abc`. On PostgreSQL 16+, the scan now uses `fsplan->checkAsUser` directly and gets the server from `fsplan->fs_server`, so neither `rte` nor `ForeignTable` is needed on that path. The RTE search is retained only for PostgreSQL 13–15 to obtain `rte->checkAsUser`.

Validated with `make lint` and full PostgreSQL 14 and 18 builds.

theory: Got it, thanks!

theory:
Thank you!

PR #321, opened 2026-07-22
Reduce NULL guarding on IN lists where possible 

JoshDreamland:
`deparseScalarArrayOpExpr`'s cheap native `IN(...)`/`NOT IN(...)` form requires the array to be provably NULL-free. A constant list with a literal `NULL` (`x IN (1, 2, NULL)`) always fails that and falls through to the guarded `CASE`, even in a `WHERE` clause, where the only actual divergence is invisible (NULL/FALSE behave the same).

To beat this horse again: a non-NULL, non-matching probe against a NULL-containing list resolves to `FALSE` in ClickHouse where Postgres computes `NULL`, but `NULL`/`FALSE` are interchangeable in a truth context. So keep track of whether we're in a truth context so we can use the cheap form for that case.

Deliberately `IN`-only, not `NOT IN` (`NOT NULL` is still `NULL`). With `transform_null_in=0`, ClickHouse's native `IN` resolves a non-matching probe against a NULL-containing list to `FALSE` (tolerable), but native `NOT IN` resolves the same case to `TRUE`, which diverges from Postgres's `NULL` in all cases, not just value-observable ones.

Confirmed empirically against a live server:

```sql
SELECT 10 IN (1, NULL), 10 NOT IN (1, NULL)  -- transform_null_in=0
→ 0, 1
```

## Implementation

A new `deparse_expr_cxt.truth_ctx` bool mirrors the walker's `ExprTruthCtx` (`foreign_expr_walker`, `EXPR_CTX_TRUTH` specifically). It is granted (set to true) at deparse's mirrors of the walker's `EXPR_CTX_TRUTH` grants (`appendConditions`, `SubPlan` quals, aggregate `FILTER`, `CASE WHEN` conditions) and degraded to `false` everywhere else via a `nodeTag` switch in `deparseExpr`. Those context cases need to be kept in sync or we'll have guards where we don't need them (an acceptable failure mode). 

Note that the walker's third state, `EXPR_CTX_NEGATED`, can't reach a plain `ScalarArrayOpExpr` due to how the planner structures that query; verified this via `EXPLAIN`.

### DIFF
--- a/src/deparse.c
+++ b/src/deparse.c
@@ -122,6 +122,17 @@ typedef struct deparse_expr_cxt {
     bool no_sort_parens; /* determines sort group clause format */
 
     /*
+     * True while the expression being deparsed only qualifies rows (as in a
+     * filter). For our purposes, denotes NULL and FALSE are interchangeable.
+     * This is the deparse-time analog of the walker's `EXPR_CTX_TRUTH`.
+     * Set to true where the walker sets `EXPR_CTX_TRUTH` (appendConditions,
+     * SubPlan quals, aggregate FILTERs, CASE WHEN conditions), restored to
+     * false by `deparseExpr` anywhere the walker assigns either other value.
+     * Only used to reduce the number of guards emitted by queries.
+     */
+    bool truth_ctx;
+
+    /*
      * True when the statement being deparsed inlines at least one SubPlan.
      * Computed once up front (a contain_subplans scan over the tlist/quals)
      * before any emission begins, and never propagated between contexts.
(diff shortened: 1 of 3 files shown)


JoshDreamland:
I think this is robust enough; the cost of missing a case in the deparser that we handle in the walker is minimal (extra guards; oh no...) and the real risk is from forgetting to set back to false, which the code isn't doing. So this should age fine even if we somehow forget about it.

theory (approved):
Please note user-visible effects in changelog.

PR #327, opened 2026-07-29
Update import_schema_1.out for ClickHouse's new Float32 rendering

JoshDreamland:
## Summary
- `0c85edc` updated the primary `import_schema.out` for ClickHouse 26.7+'s changed Float32 min/max rendering (`1.1754942e-38` → `1.1754944e-38`, `3.4028233e38` → `3.4028235e38`), but didn't touch the PG13-17 alternate, `import_schema_1.out`, which still carries the pre-26.7 values.
- `postgres.yml` installs "latest" ClickHouse unpinned across the *entire* PG version matrix (13-19), so every job now gets the new rendering — PG13-17 jobs fail comparing against the stale alternate, while PG18-19 (which use the already-updated primary) pass.
- Confirmed by rerunning CI against `main`'s current HEAD directly (no code changes): same PG13-17 failure, proving it's a live, unrelated regression rather than anything specific to #321.
- This patches only the 3 stale hunks in `import_schema_1.out` to match what's already in the primary file; the remaining (legitimate) diff between the two — PG13-17 lacking `\d`'s "Not-null constraints:" section — is untouched.

## Test plan
- [x] `make installcheck REGRESS=import_schema` passes locally (PG18, exercises the primary file)
- [ ] CI green across the PG13-19 matrix


theory (approved):
Just update `test/expected/import_schema_1.out` please. Leave the rest as-is.

PR #328, opened 2026-08-01
Update http driver to use native format

serprex:
This has multiple benefits:
* generally better memory usage
* faster encoding/decoding
* encoding/decoding now consistent between http & binary drivers
* reduced code size
* fixes a bunch of misc issues in http driver

`fetch_size` option now deprecated as native decoder handles streaming one curl chunk at a time

Also switch to copying error messages into preallocated buffers,
while avoiding truncation splitting multibyte UTF8 character

would be even simpler if we cut `clickhouse_raw_query` in #329


test/sql/http.sql
```diff
@@ -381,8 +335,8 @@ CREATE FOREIGN TABLE ft_bytea (id text, v bytea)
 SELECT id, v, v IS NULL AS is_null, octet_length(v) FROM ft_bytea ORDER BY id;
 
 /*
- * time columns strip the exact ISO epoch date prefix. Shorter values or ones
- * with another prefix route through the input function unchanged.
+ * time columns route String values through the input function unchanged, so
```
serprex: this is result of simpliying code since we no longer need to massage CH text values into pg values, so we're able to handle DateTime conversion natively

src/option.c
```diff
@@ -138,7 +136,11 @@ clickhouse_fdw_validator(PG_FUNCTION_ARGS) {
         }
 
         if (strcmp(def->defname, "fetch_size") == 0) {
-            validate_fetch_size_option(def);
+            ereport(
+                WARNING,
+                errcode(ERRCODE_WARNING_DEPRECATED_FEATURE),
+                errmsg("option \"fetch_size\" is deprecated and ignored")
```
theory:
```suggestion
                errmsg("option \"fetch_size\" is deprecated and ignored; will be removed in a future release")
```

test/sql/http.sql
```diff
@@ -263,48 +249,21 @@ CREATE FOREIGN TABLE bad_name (
...
-) SERVER http_no_stream OPTIONS (table_name 't1', fetch_size '100');
-
-/* Query with table-level streaming override (fetch_size = 100). */
-SELECT c3 FROM ft_override_stream ORDER BY c1 LIMIT 3;
-
-/* Default (no fetch_size set) uses streaming — already tested via ft1. */
-SELECT c3 FROM ft1 ORDER BY c1 LIMIT 3;
-
-/* Negative fetch_size should be rejected. */
+/* Deprecated fetch_size options warn but remain accepted. */
```
theory:
```suggestion
/* Deprecated fetch_size option warns but remains accepted for now. */
```

test/sql/http.sql
theory: So why does the comment say an ISO timestamp is rejected? We shouldn't get one at all — unless the user maps a DateTime to Time I guess.

test/expected/http.out
```diff
@@ -223,6 +192,7 @@ DETAIL:  Remote Query: SELECT c1, c3, c4 FROM "no such database".t1 ORDER BY c3
 CONTEXT:  HTTP status code: 404
 ALTER USER MAPPING FOR CURRENT_USER SERVER http_loopback OPTIONS (ADD user 'no such user');
 SELECT c3, c4 FROM ft1 ORDER BY c3, c1;  -- should fail
+WARNING:  pg_clickhouse: SELECT version() failed (HTTP status 403): Code: 516. DB::Exception: no such user: Authentication failed: password is incorrect, or there is no user with such name. (AUTHENTICATION_FAILED)
```
theory: Hrm, this might be a little confusing.

test/expected/http_inserts.out
```diff
@@ -249,8 +249,8 @@ SELECT * FROM complex ORDER BY c1;
   1 | 2020-06-01 | 2020-06-02 10:01:02-07 | t1 | fix_t1 | low1 | 2020-06-01 | 2020-06-02 10:01:02.123
   2 | 2020-06-02 | 2020-06-03 10:01:02-07 | 5  | fix_t2 | low2 | 2020-06-02 | 2020-06-03 11:01:02.234
   3 | 2020-06-03 | 2020-06-04 10:01:02-07 | 5  | fix_t3 | low3 | 2020-06-03 | 2020-06-04 12:01:02
-  4 | 1970-01-01 | 1969-12-31 16:00:00-08 | 5  | fix_t4 | low4 | 1970-01-01 | 1970-01-01 00:00:00
-  5 | 2000-01-01 | 1999-12-31 16:00:00-08 | 5  | fix_t5 | low5 | 2000-01-01 | 2000-01-01 00:00:00
+  4 | 1970-01-01 | 1969-12-31 16:00:00-08 | 5  | fix_t4 | low4 | 1970-01-01 | 1969-12-31 16:00:00
+  5 | 2000-01-01 | 1999-12-31 16:00:00-08 | 5  | fix_t5 | low5 | 2000-01-01 | 1999-12-31 16:00:00
```
theory, line 253: I don't know how we missed this before, but happy to see the improved consistency of results.

test/expected/import_schema.out
```diff
@@ -364,10 +364,10 @@ SELECT * FROM clickhouse.arrays ORDER BY c1 LIMIT 2;
 (2 rows)
 
 SELECT * FROM clickhouse.tuples ORDER BY c1 LIMIT 2;
- c1 |    c2     |  c3.a   |  c3.b   | c4 
-----+-----------+---------+---------+----
-  0 | (0,'0',1) | {0,1,1} | {0,2,2} |  0
-  1 | (1,'1',2) | {1,1,1} | {1,2,2} |  1
```
theory, line 370: Yikes.

test/expected/json_3.out
```diff
@@ -61,12 +61,8 @@ SELECT * FROM json_bin.things ORDER BY id;
...
 DETAIL:  Remote Query: SELECT id, data FROM json_test.things ORDER BY id ASC NULLS LAST
 SELECT * FROM json_http.things ORDER BY id;
- id |                               data                               
-----+------------------------------------------------------------------
-  3 | {"id": "3", "name": "gizmo", "size": "medium", "stocked": true}
-  4 | {"id": "4", "name": "doodad", "size": "large", "stocked": false}
-(2 rows)
-
+ERROR:  pg_clickhouse: unsupported JSON serialization version 0 (set output_format_native_write_json_as_string=1)
+DETAIL:  Remote Query: SELECT id, data FROM json_test.things ORDER BY id ASC NULLS LAST
```
theory, line 65: I'm surprised to see this. Is it because 24.3 doesn't support JSON? Seems odd that it would suggest setting `output_format_native_write_json_as_string` if it doesn't support it. I assume I'm missing something.

src/http.c
```diff
@@ -123,83 +125,550 @@ ch_http_connection(ch_connection_details* details) {
...
+
+/* True for a status whose body the caller reads as an error message. */
+static bool
+error_status(long status) {
+    /* Synthetic statuses carry no server body. */
+    if (status == CH_HTTP_STATUS_CANCELED || status == CH_HTTP_STATUS_TRANSPORT_ERROR) {
+        return false;
+    }
+
+    return status > 0 && status != CH_HTTP_STATUS_OK;
```
theory: I don't suppose there are other 2xx responses. Should we check for < 300?

src/http.c
```diff
@@ -123,83 +125,550 @@ ch_http_connection(ch_connection_details* details) {
...
+    if (stream->error_msg) {
+        free(stream->error_msg);
+    }
+
+    free(stream);
+}
+
+/*
+ * ch_http_stream_begin — allocate and initialize a streaming HTTP query.
+ * Returns NULL on failure.
```
theory: Should we have a string reference to know what the error is? Or does it matter?

src/http.c
```diff
@@ -123,83 +125,550 @@ ch_http_connection(ch_connection_details* details) {
...
+     * Each HttpStream owns its easy handle, so concurrent foreign scans in
+     * subqueries or joins do not fight over one.
+     */
+    stream->curl = curl_easy_init();
+    if (!stream->curl) {
+        goto fail;
     }
 
+    /* Allocate stream buffer */
+    stream->buf = (char*)malloc(INITIAL_BUF_SIZE);
```
theory, line 559: No use of memory contexts here, eh?

src/http.c
```diff
@@ -224,24 +691,48 @@ ch_http_simple_query(ch_http_connection_t* conn, const ch_query* query) {
...
+ */
+char*
+ch_http_format_error(char* errstring) {
+    size_t n = strlen(errstring);
+
+    for (size_t i = 0; i < n; i++) {
+        if (strncmp(errstring + i, "version", 7) == 0) {
+            return pnstrdup(errstring, i - 2);
+        }
+    }
```
theory: Would `strstr()` be more efficient?

src/pglink.c
```diff
@@ -877,12 +565,75 @@ http_prepare_insert(
...
+        chc_type* coltype;
+
+        /*
+         * ClickHouse gained Time64 in 25.6 and casts it to none of the types
+         * a table holds a time of day in, so send a timestamp on the epoch
+         * date, as the TabSeparated payload did.
+         */
+        if (attr->atttypid == TIMEOID) {
+            chtype = attr->attnotnull ? "DateTime64(6, 'UTC')"
+                                      : "Nullable(DateTime64(6, 'UTC'))";
```
theory, line 624: I'd really like to use Time64 where available. Not nec this PR tho.

theory (approved):
Very nice. Left some comments and questions. Once resolved let's squash to a single commit and write a nice message.

test/sql/http.sql
serprex: we were shimmying things so that string representation would map DateTime to Time through http text. I removed that, but as a result if someone hits this coercion path by bringing their own strings then they won't get that shimmying anymore

src/http.c
serprex: It's always some allocation failure, doesn't matter. Updated comment that other errors will come over stream, not from here

src/http.c
serprex: [`processOptionsRequest`](https://github.com/ClickHouse/ClickHouse/blob/b8517f0755d699ac639bd013a83675a240774d35/src/Server/HTTPHandler.cpp#L128) returns 204. We don't hit that, but yeah might as well change our checks to accept 2xx

test/expected/json_3.out
serprex: No 24.3 fails with `could not map things.data type <Object('json')>`. This is 24.8. `output_format_native_write_json_as_string` was added in 24.10. Error suggesting setting is from `clickhouse-c`

src/http.c
serprex: Part of desire to keep http.c focused on just transport. These get mixed up with curl callbacks so we don't want longjmp in there

test/sql/http.sql
theory: Oh, okay, that seems fine.

test/expected/http.out
serprex: got rid of error message, can rely on error being raised by code which follows

src/http.c
serprex: yes, switched to `ch_http_copy_error` matching how we do error handling in pg_re2 & pg_chdb

theory:
Excellent, thank you!

PR #329, opened 2026-08-01
Deprecate clickhouse_raw_query

serprex:
Add `clickhouse_perform` for DDL whose empty result has no columns

`clickhouse_raw_query` drifted from its "raw" roots with Native format,
from back when ClickHouse TSV output could be piped into text result.
Somewhat contrived massaging native blocks into TSV

This removal will allow discarding leftover TSV parsing

### DIFF
--- a/doc/pg_clickhouse.md
+++ b/doc/pg_clickhouse.md
@@ -945,6 +945,11 @@ the PostgreSQL column when importing columns; additional types may be used in
 | UInt8      | smallint         |                               |
 | UUID       | uuid             |                               |
 
+Any column also reads into `text`, `varchar`, or another string type. The value
+takes the PostgreSQL type above, then renders through that type's output
+function. UInt64 values above the bigint maximum still error, so render those
+with the ClickHouse `toString()` function.
+
 Additional notes and details follow.
 
 ### BYTEA
@@ -956,7 +961,7 @@ to [BYTEA]. Example:
 
 ```sql
 -- Create clickHouse table with String columns.
-SELECT clickhouse_raw_query($$
+CALL clickhouse_perform('ch_srv', $$
     CREATE TABLE bytes (
 	      c1 Int8, c2 String, c3 String
     ) ENGINE = MergeTree ORDER BY (c1);
@@ -1073,6 +1078,13 @@ These functions provide the interface to query a ClickHouse database.
 
 #### `clickhouse_raw_query`
 
+> [!WARNING]
+> Deprecated: `clickhouse_raw_query()` emits a deprecation warning and will be
+> removed in the next release. Use [`clickhouse_query`](#clickhouse_query) to
+> read rows and [`clickhouse_perform`](#clickhouse_perform) to run statements
+> that return none. Both reuse a configured server's driver, credentials,
+> database, and connection cache instead of an ad-hoc connection string.
+
 ```sql
 SELECT clickhouse_raw_query(
     'CREATE TABLE t1 (x String) ENGINE = Memory',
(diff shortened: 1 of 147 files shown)


doc/pg_clickhouse.md
```diff
@@ -963,11 +963,11 @@ to [BYTEA]. Example:
 
 ```sql
 -- Create clickHouse table with String columns.
-SELECT clickhouse_raw_query($$
+SELECT FROM clickhouse_query('ch_srv', $$
     CREATE TABLE bytes (
 	      c1 Int8, c2 String, c3 String
     ) ENGINE = MergeTree ORDER BY (c1);
-$$);
+$$) AS t(x int);
```
theory: Would be nice if we could do `AS (void)` or some such.

theory (commented):
We'll need to increment the major version number.

I wonder to what extent this will break use cases?

doc/pg_clickhouse.md
serprex: I was thinking same, unfortunately postgres rejects `void`

maybe we remove the 0 column hackery & introduce a separate `clickhouse_query_void` or `clickhouse_perform`

theory:
I suggest we remove all use of it, but keep the function for one more release, but have it emit a warning that it will be dropped in the next release. Mention the same in the docs.

doc/pg_clickhouse.md
theory: +1 to `clickhouse_perform`.

doc/pg_clickhouse.md
```diff
@@ -1160,23 +1167,46 @@ SELECT * FROM clickhouse_query(
...
+
+```sql
+CALL clickhouse_perform(
+    'server',
+    'CREATE TABLE remote_table (id Int32) ENGINE = MergeTree ORDER BY id'
+);
+```
+
+Execute a statement against an already-configured foreign server and discard
+any result. Use it for statements that return no columns, such as DDL, where
```
```suggestion
any result. Use it for statements that return no rows, such as DDL, where
```

doc/pg_clickhouse.md
```diff
@@ -1160,23 +1167,46 @@ SELECT * FROM clickhouse_query(
...
+[`clickhouse_query`](#clickhouse_query) has no result shape to declare. It
+resolves the server the same way `clickhouse_query` does, reusing its `driver`,
+credentials, database, and the connection cache.
+
+As a procedure it must be invoked with [CALL], not `SELECT`, and it reports no
```
```suggestion
As a procedure it must be invoked with [CALL], not `SELECT`, and it returns no
```

doc/pg_clickhouse.md
```diff
@@ -1160,23 +1167,46 @@ SELECT * FROM clickhouse_query(
...
+rows. No role has `EXECUTE` access by default:
```
```suggestion
rows. No role has `EXECUTE` access by default; `GRANT` to a role to allow it to
use the procedure.
```

sql/pg_clickhouse--0.3--0.4.sql
```diff
@@ -13,5 +13,13 @@ RETURNS SETOF record
...
--- As with clickhouse_raw_query(), don't let PUBLIC run arbitrary remote queries.
+-- Don't let PUBLIC run arbitrary remote queries.
 REVOKE EXECUTE ON FUNCTION clickhouse_query(text, text) FROM PUBLIC;
+
+-- Execute statement against foreign server, discarding any result, e.g.
+-- CALL clickhouse_perform('srv', 'CREATE TABLE t (a Int32) ENGINE = Memory');
+CREATE PROCEDURE clickhouse_perform(TEXT, TEXT)
+AS 'MODULE_PATHNAME'
+LANGUAGE C;
+
```
theory, line 24:
```suggestion

-- Don't let PUBLIC run arbitrary remote queries.
```

sql/pg_clickhouse.sql
```diff
@@ -36,9 +39,18 @@ RETURNS SETOF record
...
+-- Make sure PUBLIC can't run arbitrary remote queries; roles must be granted
+-- explicit access.
 REVOKE EXECUTE ON FUNCTION clickhouse_query(text, text) FROM PUBLIC;
 
+-- Execute statement against foreign server, discarding any result, e.g.
+-- CALL clickhouse_perform('srv', 'CREATE TABLE t (a Int32) ENGINE = Memory');
+CREATE PROCEDURE clickhouse_perform(TEXT, TEXT)
+AS 'MODULE_PATHNAME'
+LANGUAGE C;
+
```
theory, line 51:
```suggestion

-- Make sure PUBLIC can't run arbitrary remote queries; roles must be granted
-- explicit access.
```

doc/pg_clickhouse.md
```diff
@@ -1160,23 +1167,46 @@ SELECT * FROM clickhouse_query(
...
+named in the column definition list. It reuses the server's `driver`,
+credentials, database, and the connection cache.
 
 The first argument is the name of a server created with [CREATE SERVER]. A
 column definition list (`AS name(col type, ...)`) is required: PostgreSQL needs
 the result shape before fetching rows, and it must match the columns the query
 returns. Values are converted from ClickHouse to the declared types the same way
-a foreign table column would be.
+a foreign table column would be. Statements that return no columns, such as DDL,
+have nothing to declare; run them with
```
```suggestion
a foreign table column would be. Statements that return no columns, such as DDL
and `UPDATE` without `RETURNING`, which have nothing to declare; run them with:
```

src/fdw.c
```diff
@@ -525,6 +536,45 @@ clickhouse_query(PG_FUNCTION_ARGS) {
...
+    server_name = text_to_cstring(PG_GETARG_TEXT_P(0));
+    sql         = text_to_cstring(PG_GETARG_TEXT_P(1));
+    server      = GetForeignServerByName(server_name, false);
+    user        = GetUserMapping(GetUserId(), server->serverid);
+    sconn       = chfdw_get_scan_connection(user);
+
+    ch_query query = new_query(sql, 0, NULL, NULL, NULL);
+
+    /* deleting the cursor context drains the connection so it stays reusable */
+    cursor = sconn.gate.methods->simple_query(sconn.gate.conn, &query);
```
theory, line 571: Damn, that was easy! 😂

theory (approved):
This change makes me happy: huge improvement to the semantics. Using a procedure here is the prefect solution. LFG!

Mostly leaving a few copy edit nitpicks and enjoying the sea of red. 

Would you be willing to write a more descriptive commit message (50/72 format)?

doc/pg_clickhouse.md
serprex: CH lacks `RETURNING` (https://clickhouse.com/docs/reference/statements/update), & they have that whole [`ALTER TABLE ... UPDATE ...`](https://clickhouse.com/docs/get-started/quickstarts/mutations) thing

doc/pg_clickhouse.md
serprex: Going with "no results" since it avoids the whole columns/rows thing

PR #330, opened 2026-08-02
preserve typmod in binary driver

serprex:
matches http driver handling of `char(n)` columns

split from #328


serprex:
not sure if this needs CHANGELOG entry, "Preserve typmod in binary driver" seemed a bit opaque as an entry

theory (approved):
Yes please mention in the changelog. Explain in a sentence how it's user-visible, something like

> Fixed omission of typemod behavior in the binary driver so that resulting values are correct. For example, padding will be preserved in `char(x)` columns, where previously it was omitted.

PR #332, opened 2026-08-03
Preserve PostgreSQL substring offset semantics

fallintoplace:
## Summary

`pg_clickhouse` translates PostgreSQL substring functions to ClickHouse for remote execution:

| PostgreSQL overload | ClickHouse function |
| --- | --- |
| `substring(text, ...)` / `substr(text, ...)` | `substringUTF8` |
| `substring(bytea, ...)` / `substr(bytea, ...)` | `substring` |

Those mappings are only semantically safe for a subset of PostgreSQL bounds. This change retains pushdown for safe constant bounds and evaluates every other case locally in PostgreSQL.

## PostgreSQL and ClickHouse differ

The difference affects both mappings, not just text values:

| Type | PostgreSQL expression | PostgreSQL result | ClickHouse behavior before this change |
| --- | --- | --- | --- |
| `text` | `substring('val1' FROM -1)` | `val1` | A negative start counts from the end |
| `text` | `substring('val1' FROM 0 FOR 2)` | `v` | Empty result |
| `text` | `substring('val1' FROM 2 FOR -1)` | Error: `negative substring length not allowed` | Returns a shortened value |
| `bytea` | `substring('val1'::bytea FROM -1 FOR 3)` | `v` | A negative start counts from the end |
| `bytea` | `substring('val1'::bytea FROM 0 FOR 2)` | `v` | Empty result |
| `bytea` | `substring('val1'::bytea FROM 2 FOR -1)` | Error: `negative substring length not allowed` | Returns a shortened value |

Unconditionally shipping these calls changes results and can suppress a PostgreSQL error.

## Approach

Push down `substring` and `substr` only when their bounds are known-safe constants at planning time:

| Function form | Pushed down when | Evaluated locally when |
| --- | --- | --- |
| `substring(value, start)` / `substr(value, start)` | `start` is a non-NULL constant `>= 1` | `start` is NULL, non-constant, zero, or negative |
| `substring(value, start, length)` / `substr(value, start, length)` | `start >= 1` and `length >= 0`, both non-NULL constants | Either bound is NULL, non-constant, or outside that range |

The guard is applied consistently to text and bytea overloads. Safe cases such as `substring(val, 1, 3)` and `substr(val, 2)` continue to use ClickHouse. A parameterized offset is evaluated locally because its value is unavailable when the remote expression is planned.

## Tests

Regression coverage includes:

- text values with negative and zero starts
- bytea values with a negative start
- PostgreSQL's negative-length error
- a generic prepared plan with a parameterized offset
- `EXPLAIN` output confirming unsafe calls stay out of the remote SQL

Verified with:

```sh
make TESTS=test/sql/functions.sql PGUSER=postgres installcheck
```

on PostgreSQL 19 against ClickHouse 26.3.17.56 and 23.3.22.3.

### DIFF
--- a/src/custom_types.c
+++ b/src/custom_types.c
@@ -489,14 +489,16 @@ lookup_builtin_func(Oid funcid, builtin_func_def* def) {
     case F_SUBSTR_TEXT_INT4:
     case F_SUBSTRING_TEXT_INT4_INT4:
     case F_SUBSTRING_TEXT_INT4:
-        def->ch_name = "substringUTF8";
+        def->cf_type = CF_SUBSTRING;
+        def->ch_name = "\1";
         return true;
         /* PG substring(text, ...) counts code points */
     case F_SUBSTR_BYTEA_INT4_INT4:
     case F_SUBSTR_BYTEA_INT4:
     case F_SUBSTRING_BYTEA_INT4_INT4:
     case F_SUBSTRING_BYTEA_INT4:
-        def->ch_name = "substring";
+        def->cf_type = CF_SUBSTRING;
+        def->ch_name = "\1";
         return true;
         /* bytea variant is byte-based; CH substring matches */
     case F_REGEXP_LIKE_TEXT_TEXT:
--- a/src/deparse.c
+++ b/src/deparse.c
@@ -277,6 +277,8 @@ static void
 deparseArrayExpr(ArrayExpr* node, deparse_expr_cxt* context);
 static void
 deparseArrayList(ArrayExpr* node, deparse_expr_cxt* context);
+static char*
+deparseExprToString(Expr* expr, deparse_expr_cxt* context);
 static void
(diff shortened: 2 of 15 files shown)


src/shipable.c
```diff
@@ -291,6 +291,34 @@ chfdw_is_shippable(
...
+                FuncExpr* fn = (FuncExpr*)node;
+                Expr* start;
+                Const* start_const;
+
+                if (list_length(fn->args) != 2 && list_length(fn->args) != 3) {
+                    return false;
+                }
+
+                start = (Expr*)lsecond(fn->args);
+                if (!IsA(start, Const)) {
```
serprex: can we push down conditional logic to match postgres instead?

cc @JoshDreamland who recently did this sort of thing to get correct NULL handling with `IN` in #317

PR #334, opened 2026-08-03
Preserve array_position not-found semantics

fallintoplace:
## Summary

Preserves PostgreSQL `array_position()` not-found and start-position semantics for supported expressions pushed to ClickHouse.

## Change

- The two-argument form uses `nullIf(indexOf(...), 0)` so an absent value returns `NULL`.
- Positive constant third arguments use `arraySlice()` and restore the original position with `start - 1`.
- Dynamic, NULL, or non-positive third arguments remain local.
- Floating-point arrays remain eligible for pushdown. ClickHouse currently differs from PostgreSQL for `NaN` matching: `indexOf([1.1, nan], nan)` returns `0` instead of `1`. This is tracked in [ClickHouse#113169](https://github.com/ClickHouse/ClickHouse/issues/113169).
- The translated expression is parenthesized so nested arithmetic keeps PostgreSQL operator precedence.

## Tests

- Found and absent values, including `= 0` and `IS NULL` predicates.
- Positive starts, including a value before the start position, absent values, out-of-range starts, non-positive starts, and dynamic starts.
- Nested arithmetic with a three-argument call.
- `gmake installcheck` on PostgreSQL 19 / ClickHouse 26.3.
- `gmake installcheck REGRESS=array_functions` on PostgreSQL 19 / ClickHouse 23.3.

### DIFF
--- a/src/shipable.c
+++ b/src/shipable.c
@@ -248,6 +248,25 @@ chfdw_is_shippable(
                     return false;
                 }
             } break;
+            case CF_ARRAY_POSITION: {
+                FuncExpr* func = (FuncExpr*)node;
+                List* args     = func->args;
+
+                if (list_length(args) == 2) {
+                    break;
+                }
+
+                if (list_length(args) != 3) {
+                    return false;
+                }
+                /* We allow positive third argument > 0. */
+                Expr* start = (Expr*)list_nth(args, 2);
+                if (!IsA(start, Const) || ((Const*)start)->constisnull ||
+                    DatumGetInt32(((Const*)start)->constvalue) <= 0) {
+                    return false;
+                }
+                break;
+            }
             case CF_TO_CHAR: {
                 /* format must be a constant with exact CH translation */
                 Expr* fmt = (Expr*)list_nth(((FuncExpr*)node)->args, 1);
--- a/test/sql/array_functions.sql
+++ b/test/sql/array_functions.sql
@@ -17,7 +17,6 @@ SELECT clickhouse_raw_query($$
         (2, [40,50],    ['d','e'], 'x//z'),
         (3, [60],       ['f'], 'Edit -> Insert -> Line Break')
 $$);
-
 CREATE SCHEMA arr_test;
 IMPORT FOREIGN SCHEMA arr_test FROM SERVER arr_svr INTO arr_test;
 SET search_path = arr_test, public;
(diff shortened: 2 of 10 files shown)


src/shipable.c
```diff
@@ -248,6 +248,12 @@ chfdw_is_shippable(
                     return false;
                 }
             } break;
+            case CF_ARRAY_POSITION:
+                /* ClickHouse indexOf() has no start-position argument. */
```
theory: But we could mimic it by using [arraySlice](https://clickhouse.com/docs/reference/functions/regular-functions/array-functions#arraySlice), no?

test/sql/array_functions.sql
```diff
@@ -65,11 +65,22 @@ EXPLAIN (VERBOSE, COSTS OFF)
...
+-- array_position → nullIf(indexOf, 0)
 EXPLAIN (VERBOSE, COSTS OFF)
 SELECT * FROM t1 WHERE array_position(vals, 20) = 2;
 SELECT * FROM t1 WHERE array_position(vals, 20) = 2;
 
+-- array_position returns NULL rather than zero when the value is absent.
+EXPLAIN (VERBOSE, COSTS OFF)
+SELECT id FROM t1 WHERE array_position(vals, 999) IS NULL ORDER BY id;
+SELECT id FROM t1 WHERE array_position(vals, 999) IS NULL ORDER BY id;
+SELECT id FROM t1 WHERE array_position(vals, 999) = 0 ORDER BY id;
```
theory, line 77:
```suggestion
-- Returns no records because result is `NULL` not `0`.
SELECT id FROM t1 WHERE array_position(vals, 999) = 0 ORDER BY id;
```

theory (changes requested):

serprex (approved):

src/shipable.c
```diff
@@ -248,6 +250,34 @@ chfdw_is_shippable(
...
+                FuncExpr* func = (FuncExpr*)node;
+                List* args     = func->args;
+                Expr* array    = (Expr*)linitial(args);
+                Oid element    = get_base_element_type(exprType((Node*)array));
+
+                if (OidIsValid(element)) {
+                    element = getBaseType(element);
+                }
+
+                if (element == FLOAT4OID || element == FLOAT8OID) {
```
theory: What's wrong with floating point arrays?

test/sql/array_functions.sql
```diff
@@ -65,11 +78,63 @@ EXPLAIN (VERBOSE, COSTS OFF)
...
+SELECT id FROM t1 WHERE array_position(vals, 999) IS NULL ORDER BY id;
+SELECT id FROM t1 WHERE array_position(vals, 999) IS NULL ORDER BY id;
+-- Returns no records because result is `NULL` not `0`.
+SELECT id FROM t1 WHERE array_position(vals, 999) = 0 ORDER BY id;
+
+EXPLAIN (VERBOSE, COSTS OFF)
+SELECT * FROM t1 WHERE array_position(vals, 20, 2) = 2;
+SELECT * FROM t1 WHERE array_position(vals, 20, 2) = 2;
+
+EXPLAIN (VERBOSE, COSTS OFF)
```
theory, line 84:
```suggestion
-- Third argument constant > 0 should push down arraySlice().
EXPLAIN (VERBOSE, COSTS OFF)
```

test/sql/array_functions.sql
```diff
@@ -65,11 +78,63 @@ EXPLAIN (VERBOSE, COSTS OFF)
...
+SELECT id FROM t1
+WHERE array_position(vals, 20, 2) * 10 = 20
+ORDER BY id;
+SELECT id FROM t1
+WHERE array_position(vals, 20, 2) * 10 = 20
+ORDER BY id;
+
+EXPLAIN (VERBOSE, COSTS OFF)
+SELECT id FROM t1
```
theory, line 94:
```suggestion
-- Third argument should push down arraySlice(), not found should return NULL.
SELECT id FROM t1
```

test/sql/array_functions.sql
```diff
@@ -65,11 +78,63 @@ EXPLAIN (VERBOSE, COSTS OFF)
...
+WHERE array_position(vals, 999, 2) IS NULL
+ORDER BY id;
+SELECT id FROM t1
+WHERE array_position(vals, 999, 2) IS NULL
+ORDER BY id;
+
+SELECT id FROM t1
+WHERE array_position(vals, 20, 100) IS NULL
+ORDER BY id;
```
theory, line 104: Let's also test `array_position(vals, 10, 2) IS NULL` to confirm it doesn't find index 1 when we start the slice at 2.

test/sql/array_functions.sql
```diff
@@ -65,11 +78,63 @@ EXPLAIN (VERBOSE, COSTS OFF)
...
+
+EXPLAIN (VERBOSE, COSTS OFF)
```
theory:
```suggestion
-- Third argument < 1 should prevent pushdown.
EXPLAIN (VERBOSE, COSTS OFF)
```

test/sql/array_functions.sql
```diff
@@ -65,11 +78,63 @@ EXPLAIN (VERBOSE, COSTS OFF)
...
+SELECT id FROM t1
+WHERE array_position(vals, 20, 0) = 2
+ORDER BY id;
+
+EXPLAIN (VERBOSE, COSTS OFF)
```
theory, line 112:
```suggestion
-- Non-constant third argument should not push down.
EXPLAIN (VERBOSE, COSTS OFF)
```

test/sql/array_functions.sql
```diff
@@ -65,11 +78,63 @@ EXPLAIN (VERBOSE, COSTS OFF)
...
+SELECT id FROM t1
+WHERE array_position(vals, 20, id) IS NOT NULL
+ORDER BY id;
+
+EXPLAIN (VERBOSE, COSTS OFF)
+SELECT id FROM float_arrays
```
theory: I'm not sure we should prevent pushdown for float arrays; most people won't be using `NaN`. I think we should keep it. This seems like a ClickHouse bug TBH:

```
77c2a6ba3a52 :) select indexOf([1.1, nan], nan);
   ┌─indexOf([1.1, nan], nan)─┐
1. │                        0 │
   └──────────────────────────┘
```

Filed ClickHouse/ClickHouse#113169

theory (changes requested):

src/shipable.c
fallintoplace: I checked ClickHouse’s behavior while implementing this and found that indexOf([1.1, nan], nan) returns 0, while PostgreSQL’s array_position() returns 1. That’s why I initially kept floating-point arrays local.

test/sql/array_functions.sql
fallintoplace: Thank you, I have reverted this change for now. I will be keeping an ear on what the resolution is.

theory (approved):
Thank you, this is great!

theory:
Can use `arrayFirstIn()` with float arrays to properly evaluate `nan`, ahtough it requires evaluating the expression twice: Once for `x = y` and once for `isNaN(y)`:

```
 select arrayFirstIndex(x -> (x = nan) OR (isNaN(x) AND isNaN(nan)), [1.1, nan])

   ┌─arrayFirstIn⋯[1.1, nan])─┐
1. │                        2 │
   └──────────────────────────┘
```

theory:
This is better:

```
173075546ba8 :) select arrayFirstIndex(x -> (x in (nan)), [1.1, nan]);
   ┌─arrayFirstIn⋯[1.1, nan])─┐
1. │                        2 │
   └──────────────────────────┘
```

So for float arrays, we can use `arrayFirstIndex(x -> (x in (y)), array)`.

PR #335, opened 2026-08-03
Reject JSON paths containing NULL elements

fallintoplace:
## Summary

Prevents `json_extract_path*()` and `jsonb_extract_path*()` calls with a `NULL` path element from being pushed down.

## Problem

PostgreSQL returns `NULL` when any element of the variadic path is `NULL`. The FDW accepted such paths, then deconstructed the `text[]` and converted every element to text without consulting its null bitmap. That could dereference a NULL datum during planning or deparsing.

## Change

Reject constant paths containing a `NULL` element in the existing JSON-path shippability gate. PostgreSQL evaluates those expressions locally; non-NULL constant paths retain dot-notation pushdown.

## Tests

- Added planner regressions for text-returning and JSON-returning path functions.
- `gmake installcheck REGRESS=json` on PostgreSQL 19 with ClickHouse 26.3 and 23.3.
- `gmake installcheck` on PostgreSQL 19 with ClickHouse 26.3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,6 +130,9 @@ All notable changes to this project will be documented in this file. It uses the
     ClickHouse does not find an element. Thanks to Minh Vu for the PR ([#334])!
 *   Fixed `array_length()` pushdown to preserve empty-array and requested
     dimension semantics. Thanks to Minh Vu for the PR ([#336])!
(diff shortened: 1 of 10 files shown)


theory:
Quite sensible, thank you. I wrote a more descriptive commit message and linked to the PR in the change log.

theory (approved):

PR #337, opened 2026-08-03
Avoid NULL dereference for ordered aggregates

fallintoplace:
## Summary

`chfdw_check_for_ordered_aggregate()` called `get_extension_name()` and then `strcmp()` without handling aggregates that are not owned by an extension. Planning an extensionless ordered-set aggregate could therefore dereference NULL and crash the backend.

This change returns false for invalid or missing extension OIDs, frees the extension name, and adds a regression covering a local ordered-set aggregate. The regression confirms that the aggregate stays local while the foreign scan retrieves its input column.

## Testing

- `pg-build-test`
- `make installcheck REGRESS=functions`

### DIFF
--- a/src/custom_types.c
+++ b/src/custom_types.c
@@ -345,10 +345,8 @@ chfdw_check_for_ordered_aggregate(Aggref* agg) {
     }
 
     /* Accept all ordered aggregates defined by pg_clickhouse. */
-    Oid extoid    = getExtensionOfObject(ProcedureRelationId, agg->aggfnoid);
-    char* extname = get_extension_name(extoid);
-
-    return STR_EQUAL(extname, "pg_clickhouse");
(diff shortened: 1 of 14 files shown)


src/custom_types.c
```diff
@@ -345,9 +345,20 @@ chfdw_check_for_ordered_aggregate(Aggref* agg) {
...
 
-    return STR_EQUAL(extname, "pg_clickhouse");
+    if (!OidIsValid(extoid)) {
+        return false;
+    }
+
+    extname = get_extension_name(extoid);
+    result  = extname != NULL && STR_EQUAL(extname, "pg_clickhouse");
+    if (extname != NULL) {
+        pfree(extname);
```
serprex: tempted to skip `get_extension_name`'s `pstrdup` & just compare against cached extname

serprex (approved):

theory (approved):

theory:
Thank you!

PR #338, opened 2026-08-03
Preserve fractional seconds in to_timestamp()

fallintoplace:
## Summary

to_timestamp(float8) was pushed down as fromUnixTimestamp(toInt64(...)), which truncated fractional Unix timestamps. Leave the built-in for PostgreSQL evaluation so subsecond precision is preserved.

The regression coverage checks that 2042323443.232 does not match 2042323443, and does match the fractional value.

## Testing

- The extension builds and installs with -Werror.
- The to_timestamp section of make installcheck REGRESS=functions matches.
- This ClickHouse 26.7 image still has unrelated existing base64/base64url, JSON, and timeout expectation differences.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -150,6 +150,9 @@ All notable changes to this project will be documented in this file. It uses the
 *   Taught `json_extract_path*()` and `jsonb_extract_path*()` not to push down
     when the path array contains any `NULL` elements. Thanks to Minh Vu for
     the PR ([#335])!
+*   Fixed loss of subsecond precision in `to_timestamp(float8)` by mapping it
+    to the ClickHouse `toDateTime64()` function instead of
+    `fromUnixTimestamp()`. Thanks to Minh Vu for the PR ([#338]).
 
 ### 📚 Documentation
 
@@ -222,6 +225,8 @@ All notable changes to this project will be documented in this file. It uses the
     "ClickHouse/pg_clickhouse#335 Reject JSON paths containing NULL elements"
   [#336]: https://github.com/ClickHouse/pg_clickhouse/pull/336
     "ClickHouse/pg_clickhouse#336 Preserve array_length semantics"
+  [#338]: https://github.com/ClickHouse/pg_clickhouse/pull/338
+    "ClickHouse/pg_clickhouse#338 Preserve fractional seconds in to_timestamp()"
 
 ## [v0.3.2] — 2026-06-16
 
--- a/test/sql/functions.sql
+++ b/test/sql/functions.sql
@@ -374,6 +374,7 @@ EXPLAIN (VERBOSE, COSTS OFF) SELECT * FROM t6 WHERE to_timestamp(i64) = to_times
 SELECT * FROM t6 WHERE to_timestamp(i64) = to_timestamp(2042323443);
 EXPLAIN (VERBOSE, COSTS OFF) SELECT * FROM t6 WHERE to_timestamp(f64) = to_timestamp(2042323443);
 SELECT * FROM t6 WHERE to_timestamp(f64) = to_timestamp(2042323443);
+SELECT * FROM t6 WHERE to_timestamp(f64) = to_timestamp(2042323443.232);
 
 -- Check current_*-type functions.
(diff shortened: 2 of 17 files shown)


theory:
This is not the correct solution. Switch to `toDateTime64()` instead.

Postgres:

```
david=# select to_timestamp(1284352323.32300) ;
        to_timestamp        
----------------------------
 2010-09-13 04:32:03.323+00
(1 row)
```

ClickHouse:

```
:) set date_time_output_format='iso'

:) select toDateTime64(1284352323.32300, 6, 'UTC');

   ┌─toDateTime64(⋯23, 6, 'UTC')─┐
1. │ 2010-09-13T04:32:03.323000Z │
   └─────────────────────────────┘
```

test/sql/functions.sql
```diff
@@ -363,14 +363,14 @@ EXPLAIN (VERBOSE, COSTS OFF) SELECT key1, val FROM t3_map WHERE md5(val) LIKE 'a
 SELECT key1 FROM t3_map WHERE md5(val) LIKE 'a%' ORDER BY key1;
 
 -- Check to_timestamp(float8).
-EXPLAIN (VERBOSE, COSTS OFF) SELECT * FROM t6 WHERE to_timestamp(i64) = to_timestamp(0);
+EXPLAIN (VERBOSE, COSTS OFF) SELECT * FROM t6 WHERE to_timestamp(i64) IS NOT NULL;
 SELECT * FROM t6 WHERE to_timestamp(i64) = to_timestamp(0);
-EXPLAIN (VERBOSE, COSTS OFF) SELECT * FROM t6 WHERE to_timestamp(f64) = to_timestamp(0);
+EXPLAIN (VERBOSE, COSTS OFF) SELECT * FROM t6 WHERE to_timestamp(f64) IS NOT NULL;
```
theory: Are these not coming out the same? We want to make sure they're equivalent, not just non-NULL.

CHANGELOG.md
```diff
@@ -95,6 +95,8 @@ All notable changes to this project will be documented in this file. It uses the
 *   Fixed a use-after-free of a foreign scan's batch memory context on rescan
     that could corrupt memory and hang ([#296]).
 *   Fixed subsecond precision lost inserting timestamps over HTTP ([#300]).
+*   Fixed loss of subsecond precision in `to_timestamp(float8)` by mapping it
+    to ClickHouse's microsecond-precision `toDateTime64()` function.
```
theory: Link to this PR.

CHANGELOG.md
```diff
@@ -601,7 +603,7 @@ pg_clickhouse v0.1 will get its benefits on reload without needing to
 ### ⚡ Improvements
 
 *   Added support for the PostgreSQL `to_timestamp(float8)` function, mapped to
-    `fromUnixTimestamp(toInt64())` in ClickHouse.
+    `toDateTime64(value, 6, 'UTC')` in ClickHouse.
```
theory: Do not change this

test/sql/functions.sql
```diff
@@ -363,14 +363,14 @@ EXPLAIN (VERBOSE, COSTS OFF) SELECT key1, val FROM t3_map WHERE md5(val) LIKE 'a
...
 SELECT * FROM t6 WHERE to_timestamp(f64) = to_timestamp(0);
-EXPLAIN (VERBOSE, COSTS OFF) SELECT * FROM t6 WHERE to_timestamp(i64) = to_timestamp(0);
```
theory, line 373: We still need to be sure that epoch 0 is the same. Please restore.

theory (changes requested):
Much better, thanks. Just needs a few tweaks.

test/sql/functions.sql
fallintoplace: Thanks, fixed the remaining tweaks. I restored the epoch-0 coverage, changed the integer and float tests to verify equality, added the PR link to the changelog, and restored the historical changelog entry. I also rebased onto `main` and pushed commit `221d64d`. 

Please take another look when you have a chance.

theory (approved):

PR #339, opened 2026-08-03
Preserve precision for numeric power functions

fallintoplace:
## Summary

ClickHouse evaluates `pow()` as Float64, so pushing down PostgreSQL `numeric` overloads can round values before an exact comparison. Keep `float8` pushdown, but evaluate numeric `pow()` and `power()` locally in PostgreSQL.

The regression uses a Decimal128 source value to cover both function names: the exact PostgreSQL result matches, while the rounded Float64 result does not.

## Testing

- Extension builds and installs with `-Werror`
- Numeric power regression verified with PostgreSQL 19 and ClickHouse 26.7

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,12 +22,17 @@ All notable changes to this project will be documented in this file. It uses the
     custom aggregates that were not part of an extension. It now properly
     keeps custom aggregate execution local. Thanks to Minh Vu for the PR
     ([#337])!
+*   Fixed precision loss when pushing down numeric `pow()` and `power()`
+    expressions; they now execute locally in PostgreSQL. Thanks to Minh Vu for
+    the PR ([#339])!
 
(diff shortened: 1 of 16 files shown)


theory (changes requested):
[ClickHouse `pow()`](https://clickhouse.com/docs/reference/functions/regular-functions/math-functions#pow) takes a Decimal value, so we should be able to evaluate numerics without loss of precision.

theory:
Oh I see, it returns Float64. I don't know whether this loss of precision matters, though.

fallintoplace:
Yes, it matters for exact comparisons. In ClickHouse 26.7, pow(Decimal128(...), 2) accepts Decimal input but returns Float64, producing 1.5241578753238834 instead of PostgreSQL’s exact numeric result 1.524157875323883675. The regression demonstrates that pushing it down changes the result, so numeric pow()/power() should remain local while float8 remains pushed down.

theory (approved):

theory:
Thank you!

PR #340, opened 2026-08-03
Preserve NULLs in array_agg pushdown

fallintoplace:
## Summary

ClickHouse groupArray() skips NULL values, while PostgreSQL array_agg() preserves them. Wrap nullable inputs in a tuple before groupArray and unwrap them with arrayMap so NULL elements survive the pushdown.

Non-nullable array_agg() plans keep their existing SQL. Ordered array_agg() stays local because the wrapper does not preserve ordering.

## Testing

- Built and installed with -Werror
- make installcheck REGRESS=aggregates

### DIFF
--- a/src/deparse.c
+++ b/src/deparse.c
@@ -324,6 +324,8 @@ deparseRangeTblRef(
 static void
 deparseAggref(Aggref* node, deparse_expr_cxt* context);
 static void
+deparseArrayAggref(Aggref* node, deparse_expr_cxt* context);
+static void
 deparseWindowFunc(WindowFunc* node, deparse_expr_cxt* context);
 static void
 appendGroupByClause(List* tlist, deparse_expr_cxt* context);
@@ -1235,7 +1237,9 @@ foreign_expr_walker(Node* node, foreign_glob_cxt* glob_cxt, ExprTruthCtx ctx) {
         }
 
         /* groupConcat has no ORDER BY; block ordered string_agg */
-        if (agg->aggfnoid == F_STRING_AGG_TEXT_TEXT && agg->aggorder != NIL) {
+        if ((agg->aggfnoid == F_STRING_AGG_TEXT_TEXT ||
+             agg->aggfnoid == F_ARRAY_AGG_ANYNONARRAY) &&
+            agg->aggorder != NIL) {
             return false;
         }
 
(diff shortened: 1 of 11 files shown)


src/include/fdw.h
```diff
@@ -435,6 +434,7 @@ typedef enum {
                                 * length(arr)-n) */
     CF_ARRAY_SORT_DESC,        /* array_sort(arr,desc) →
                                 * arrayReverseSort/arraySort */
+    CF_ARRAY_AGG,
```
theory: Needs explanatory comment

src/deparse.c
```diff
@@ -1214,7 +1216,9 @@ foreign_expr_walker(Node* node, foreign_glob_cxt* glob_cxt, ExprTruthCtx ctx) {
         }
 
         /* groupConcat has no ORDER BY; block ordered string_agg */
-        if (agg->aggfnoid == F_STRING_AGG_TEXT_TEXT && agg->aggorder != NIL) {
+        if ((agg->aggfnoid == F_STRING_AGG_TEXT_TEXT ||
+             agg->aggfnoid == F_ARRAY_AGG_ANYNONARRAY) &&
+            agg->aggorder != NIL) {
```
theory, line 1242: Need test rejecting use of ORDER with F_ARRAY_AGG_ANYNONARRAY

test/sql/aggregates.sql
```diff
@@ -240,6 +250,11 @@ SELECT array_agg(cost) FROM agg_bin.hits WHERE id < :id_limit;
 EXPLAIN (VERBOSE, COSTS OFF) SELECT array_agg(cost) FROM agg_http.hits WHERE cost < :id_limit;
 SELECT array_agg(cost) FROM agg_http.hits WHERE id < :id_limit;
 
+EXPLAIN (VERBOSE, COSTS OFF) SELECT array_agg(v) FROM agg_bin.null_agg;
+SELECT array_agg(v) FROM agg_bin.null_agg;
+EXPLAIN (VERBOSE, COSTS OFF) SELECT array_agg(v) FROM agg_http.null_agg;
+SELECT array_agg(v) FROM agg_http.null_agg;
```
theory, line 256: This is useful but we also need tests to cover all of the variants in `deparseArrayAggref()`: filter, distinct, and nullable.

theory (changes requested):
Good find. Please add a changelog item and write a descriptive commit 72/50 message in a single commit.

PR #341, opened 2026-08-04
Preserve NaN semantics for floating-point array_position pushdown

fallintoplace:
## Overview

Fix `array_position()` pushdown for `real[]` and `double precision[]` columns.

ClickHouse `indexOf()` can miss a `NaN` when the array value and the search value were produced independently. PostgreSQL treats those values as a match for `array_position()`, so the existing pushdown could return `NULL` incorrectly.

## Implementation

- Keep the existing `nullIf(indexOf(...), 0)` path for non-floating-point arrays.
- Use `arrayFirstIndex()` with an explicit predicate for floating-point arrays.
- Match ordinary values, `NULL`, and any `NaN` payload.
- Qualify captured columns with the generated relation alias so lambda parameters cannot shadow outer columns such as `x`.
- Emit `NaN`, `Infinity`, and `-Infinity` constants as ClickHouse numeric float literals.
- Preserve positive `start` handling with `arraySlice()` and restore the original PostgreSQL position.

The floating-point predicate is equivalent to:

```sql
nullIf(
    arrayFirstIndex(
        value -> (
            (isNull(value) AND isNull(r1.needle))
            OR (value = r1.needle)
            OR (isNaN(value) AND isNaN(r1.needle))
        ),
        r1.float_array
    ),
    0
)
```

## Regression coverage

- Float32 and Float64 arrays
- Ordinary matches and missing values
- Independently produced NaNs with different payloads
- A captured column named `x`, matching the lambda parameter name
- `NaN`, positive infinity, and negative infinity constants
- Positive start positions and empty arrays
- Nullable elements and `NULL` needles, including exact returned positions
- Generated remote SQL and relation qualification

## Related issue

[ClickHouse/ClickHouse#113169](https://github.com/ClickHouse/ClickHouse/issues/113169) tracks the underlying floating-point `NaN` matching behavior in ClickHouse.

## Follow-up fix

The follow-up commit also covers array_position() expressions used only in pushed-down ORDER BY pathkeys. These expressions are resolved after the base relation is emitted, so they now share the same pathkey resolver during alias preflight and SQL generation. This keeps captured columns qualified in cases such as ORDER BY array_position(float64_vals, x), id.

Added regression coverage for the ORDER BY path and qualified generated SQL.


src/deparse.c
```diff
@@ -3462,11 +3524,20 @@ deparseConst(Const* node, deparse_expr_cxt* context, int showtype) {
...
+            strcmp(extval, "NaN") == 0) {
+            appendStringInfoString(buf, "nan");
+        } else if (
+            (node->consttype == FLOAT4OID || node->consttype == FLOAT8OID) &&
+            (strcmp(extval, "Infinity") == 0 || strcmp(extval, "+Infinity") == 0)
+        ) {
+            appendStringInfoString(buf, "inf");
+        } else if (
+            (node->consttype == FLOAT4OID || node->consttype == FLOAT8OID) &&
+            strcmp(extval, "-Infinity") == 0
```
theory: Use case-insensitive comparisons

theory (changes requested):
Partial review, more later.

PR #342, opened 2026-08-06
Refactor, mostly binary driver

serprex:
pg-clickhouse-c & native http changes left room for cleanup,
moved fdw logic from src/binary to pglink.c, maybe pglink.c needs to be broken up


src/endpoint.c
```diff
@@ -0,0 +1,66 @@
...
+/* These are the default ports for the native protocol. */
+#define CH_NATIVE_PORT 9000
+#define CH_NATIVE_TLS_PORT 9440
+
+static int
+ends_with(const char* s, const char* suffix) {
+    size_t slen       = strlen(s);
+    size_t suffix_len = strlen(suffix);
+
+    return suffix_len <= slen && !strcmp(s + slen - suffix_len, suffix);
```
theory, line 21: I wonder if we should make it case-insensitive.

theory (approved):

src/endpoint.c
serprex: I think it's good to leave it, people really shouldn't be passing uppercase into domain names

PR #344, opened 2026-08-08
Add pushdown for JSONB ? operator, with casting

theory:
Added pushdown for the JSONB `?` existence operator. Evaluates not only ClickHouse JSON columns mapped to JSONB, but also ClickHouse String columns that contain JSON, thus working on both JSON objects and JSON arrays. This requires a bit of complicated ClickHouse function dispatch, but does the trick.

Test output varies on ClickHouse 24.3 and earlier because the "JSON" feature at that time required the same keys in every object in the column, so the tests do in fact find the key in every row.

Resolves #325

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,10 @@ All notable changes to this project will be documented in this file. It uses the
     go binary encoding, the HTTP driver always streams results the same as the
     binary driver. Setting `fetch_size` triggers a warning and will be removed
     in a future release. ([#328])
+*   Added pushdown for the JSONB `?` existence operator. Evaluates not only
+    ClickHouse JSON columns mapped to JSONB, but also ClickHouse String
+    columns that contain JSON, thus working on both JSON objects and JSON
+    arrays. Thanks to Kostia R for the PR ([#344])!
 
 ### 🐞 Bug Fixes
 
@@ -203,6 +207,8 @@ All notable changes to this project will be documented in this file. It uses the
     "ClickHouse/pg_clickhouse#317 Push down the array IN family unconditionally"
   [#319]: https://github.com/ClickHouse/pg_clickhouse/pull/319
     "ClickHouse/pg_clickhouse#319 Fix foreign scan RTE selection"
+  [#344]: https://github.com/ClickHouse/pg_clickhouse/pull/344
+    "ClickHouse/pg_clickhouse#344 Push down JSONB existence predicates"
   [#326]: https://github.com/ClickHouse/pg_clickhouse/pull/326
     "ClickHouse/pg_clickhouse#326 pg-clickhouse-c"
   [#328]: https://github.com/ClickHouse/pg_clickhouse/pull/328
--- a/src/deparse.c
+++ b/src/deparse.c
@@ -198,6 +198,12 @@ typedef enum ExprTruthCtx {
     EXPR_CTX_EXACT,
 } ExprTruthCtx;
 
+typedef enum JsonbDocumentKind {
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -204,6 +207,8 @@ All notable changes to this project will be documented in this file. It uses the
     "ClickHouse/pg_clickhouse#317 Push down the array IN family unconditionally"
   [#319]: https://github.com/ClickHouse/pg_clickhouse/pull/319
     "ClickHouse/pg_clickhouse#319 Fix foreign scan RTE selection"
+  [#325]: https://github.com/ClickHouse/pg_clickhouse/pull/325
+    "ClickHouse/pg_clickhouse#325 Push down JSONB existence predicates"
```
serprex: this is #344

test/sql/jsonb_exists.sql
```diff
@@ -0,0 +1,82 @@
...
+
+SELECT clickhouse_raw_query('DROP DATABASE IF EXISTS jsonb_exists');
+SELECT clickhouse_raw_query('CREATE DATABASE jsonb_exists');
+SELECT clickhouse_raw_query($$
+    CREATE TABLE jsonb_exists.native_documents (
+        id Int32,
+        document JSON
+    ) ENGINE = MergeTree ORDER BY id
+$$);
+SELECT clickhouse_raw_query($$
```
serprex: should use `clickhouse_perform`

test/expected/jsonb_exists.out
```diff
@@ -0,0 +1,161 @@
...
+(3 rows)
+
+SELECT id FROM jsonb_exists_compatibility
+WHERE document ? 'key'
+ORDER BY id;
+ id 
+----
+  1
+  2
+(2 rows)
```
serprex, line 81: on postgres this would include `"key"` (6)

src/deparse.c
```diff
@@ -4930,6 +5019,85 @@ findFunction(Oid typoid, char* name) {
...
+    appendStringInfoString(buf, ") = 'Object', JSONHas(");
+    deparseJsonbNonnullDocument(document, document_kind, context);
+    appendStringInfoString(buf, ", ");
+    deparseExpr(key, context);
+    /* Use arrayExists() if the JSON is an array. */
+    appendStringInfoString(buf, "), JSONType(");
+    deparseJsonbNonnullDocument(document, document_kind, context);
+    appendStringInfoString(
+        buf,
+        ") = 'Array', arrayExists(jsonb_exists_element -> "
```
serprex: This will cause trouble if they have column named `jsonb_exists_element`

theory: I actually prefer `x`, but maybe it should be `__x`.

test/sql/jsonb_exists.sql
theory: Fixed in f003fc0

CHANGELOG.md
theory: Original is 325 but I take your point.

test/expected/jsonb_exists.out
theory:
```
david=# select '{"key": 1}'::jsonb ? '"key"';
 ?column? 
----------
 f
```

src/deparse.c
theory: Renamed to `elem` and added a column named `elem` to the tests that doesn't conflict AFAICT.

serprex: conflict is when you're doing `col1 ? elem`

PR #345, opened 2026-08-08
Increment to v0.10.0, add co-maintainers

theory:
And copy-edit the changelog.


serprex (approved):

JoshDreamland (commented):
Post-submit approval since checking all the Viewed boxes is not quite the same as approving (technically).

PR #346, opened 2026-08-11
remove clickhouse_raw_query

serprex:
Started unreleased 1.0 section, guessing we'll want 1.0 to be release where we switch pg_clickhouse itself to being a meta extension, not sure exact plan for executing that


doc/pg_clickhouse.md
```diff
@@ -1,5 +1,5 @@
-pg_clickhouse 0.10.0
-====================
+pg_clickhouse 1.0.0
```
theory: I suggest we wait till we have all the TPC-H queries pushing down before going to v1.0.

CHANGELOG.md
```diff
@@ -7,6 +7,17 @@ All notable changes to this project will be documented in this file. It uses the
...
     "Semantic Versioning 2.0.0"
 
+## [v1.0.0] — Unreleased
+
+### ⚡ Improvements
+
+*   Removed `clickhouse_raw_query()`, deprecated in v0.10.0. Use
+    `clickhouse_query(server, sql)` to read rows and
+    `CALL clickhouse_perform(server, sql)` to run statements that return none;
+    both take a foreign server rather than a connection string.
```
theory: Link to PR.

theory (approved):

doc/pg_clickhouse.md
serprex: adjusted to 0.11, pg_clickhouse being a metapackage should be 1.0, clickhouse_fdw 1.0 I agree with you on. & I guess most of this will end up going to clickhouse_fdw in that scenario so it'd be 0.11 there

PR #349, opened 2026-08-19
use pgch_pg_type_for in IMPORT SCHEMA

serprex:
Replace pg_clickhouse's ad hoc declaration parser with mapping built on clickhouse-c's type tree
Keeps encoding, decoding, and schema import aligned while preserving typmod, array depth, and nullability

Depends on ClickHouse/pg-clickhouse-c/pull/8


serprex:
https://github.com/ClickHouse/pg_chdb/pull/60

doc/pg_clickhouse.md
```diff
@@ -913,31 +922,49 @@ cluster to be restart when the library is updated.
...
+|-------------------|------------------|----------------------------------|
+| Array(T)          | T[]              | One PG array type per depth      |
+| Bool              | boolean          |                                  |
+| Date              | date             |                                  |
+| Date32            | date             |                                  |
+| DateTime          | timestamptz      |                                  |
+| DateTime64(P)     | timestamptz(P)   | P over 6 caps at 6               |
+| Decimal(P,S)      | numeric(P,S)     |                                  |
+| Enum8             | text             |                                  |
+| Enum16            | text             |                                  |
```
theory: In the back of my mind I've been wondering about supporting enum definitions in  `IMPORT FOREIGN SCHEMA`. Seems do-able.

doc/pg_clickhouse.md
```diff
@@ -913,31 +922,49 @@ cluster to be restart when the library is updated.
...
+| IPv4              | inet             |                                  |
+| IPv6              | inet             |                                  |
+| Int8              | smallint         |                                  |
+| Int16             | smallint         |                                  |
+| Int32             | integer          |                                  |
+| Int64             | bigint           |                                  |
+| JSON              | jsonb, json      |                                  |
+| LineString        | path             |                                  |
+| LowCardinality(T) | T                |                                  |
+| Map(K,V)          | text[]           | Import renders its pairs as text |
```
theory: so `ARRAY[[k, v], [k, v]...]`?

doc/pg_clickhouse.md
```diff
@@ -913,31 +922,49 @@ cluster to be restart when the library is updated.
...
+| MultiLineString   | path[]           |                                  |
+| MultiPolygon      | polygon[]        |                                  |
+| Nullable(T)       | T                | Column imports without NOT NULL  |
+| Point             | point            |                                  |
+| Polygon           | polygon[]        |                                  |
+| Ring              | polygon          |                                  |
+| String            | text, bytea      |                                  |
+| Time              | time             |                                  |
+| Time64(P)         | time(P)          | P over 6 caps at 6               |
+| Tuple(...)        | text             | Import renders it as text        |
```
theory: We should add a note somewhere if we haven't already that one can define a composite type. Maybe we could create them in `IMPORT FOREIGN SCHEMA`, too.

src/pglink.c
```diff
@@ -918,192 +918,250 @@ binary_finalize_insert(void* istate) {
 }
 
 /*
- * Query to generate table for doc/pg_clickhouse.md. Keep in sync with
- * str_types_map below. On change, re-run and paste the output into
- * doc/pg_clickhouse.md. Perl: https://stackoverflow.com/a/58443028/79202
+ * Query to generate table for doc/pg_clickhouse.md. Keep in sync with the
+ * column descriptors pgch_pg_type_for() reports. On change, re-run and paste
+ * the output into doc/pg_clickhouse.md.
+ * Perl: https://stackoverflow.com/a/58443028/79202
```
theory: We might want to move this to the pg-clickhouse docs, since I expect to generate a similar table for chdb.

test/expected/import_types.out
```diff
@@ -0,0 +1,82 @@
...
+CALL clickhouse_perform('import_types_admin', 'INSERT INTO import_types_test.mapped VALUES (
+    1, 1.5, NULL, [1.25, -2.5], [[1.25], [2.5]],
+    ''2026-08-19 03:04:05.678'', [''2026-08-19 03:04:05.678901''],
+    ''2026-08-19 03:04:05.678901234'',
+    ''abcde'', [''xy'', ''z''], ''two'', ''card'', {''k'': 42}, (7, ''seven'')
+)');
+CREATE SCHEMA import_types;
+IMPORT FOREIGN SCHEMA import_types_test FROM SERVER import_types_loopback
+    INTO import_types;
+NOTICE:  pg_clickhouse: ClickHouse <Map> type was translated to <TEXT> type for column "pairs", please create composite type and alter the column if needed
```
theory: Actually `text[]`.

theory (commented):
Oh this is big. Will be able to just delete the `time64` branch.

src/pglink.c
serprex: ClickHouse/pg-clickhouse-c/pull/9

doc/pg_clickhouse.md
serprex: pg-clickhouse-c wants this to come in as record[] but we can't create that in table. unless we start creating types when we import schema..

currently ends up being `{"(k,42)"}` would be nice to get this to be `text[][]`

theory: Yeah `"(k,42)"}` is not great.

doc/pg_clickhouse.md
serprex: Probably put it behind `OPTIONS (import_types = true)` on `IMPORT FOREIGN SCHEMA`. Will involve some more changes in pg-clickhouse-c so pg_chdb can share that logic. Probably best as followup

doc/pg_clickhouse.md
serprex: ClickHouse/pg-clickhouse-c/pull/10 adds support to decode Tuple into postgres array

doc/pg_clickhouse.md
```diff
@@ -249,6 +249,18 @@ the foreign tables. Columns will be defined using the [supported data
...
+below every `Array` layer, so `Array(Decimal(12,6))` imports as
+`numeric(12,6)[]`. A `Nullable` column imports without `NOT NULL`; `Nullable`
+inside an `Array` does not, because PostgreSQL arrays always allow NULL
+elements. `Tuple` and `Map` columns import as `text[]` and `text[][]` with a
+`NOTICE`, since PostgreSQL cannot hold their `record` types: a `Tuple` fills
+an array with its fields, a `Map` one row per pair. A `DateTime64` or
+`Time64` past microseconds also imports with a `NOTICE`, since its column caps
+at `timestamp(6)` or `time(6)`. The import rejects a column whose type has no
+PostgreSQL counterpart, including the legacy `Object('json')` type that
+predates ClickHouse `JSON`.
```
theory: Meant to make this a patch:

```md
Some details to keep in mind:

*   Imported columns keep the type modifier of the ClickHouse type, including
    those within `Array`s. In other words, `Array(Decimal(12,6))` imports as
    `numeric(12,6)[]`.

*   `Nullable` columns import without `NOT NULL`; `Nullable` inside an `Array`
    does not, because PostgreSQL arrays always allow NULL elements.

*   `Tuple` and `Map` columns import as `text[]` and `text[][]` with a
    `NOTICE`, since PostgreSQL cannot map them to `record` types, because they
    cannot be used as a column. Therefore, a `Tuple` fills an array with its
    fields, while a `Map` fills an array with a nested array for each key
    pair.

*   `DateTime64` or `Time64` columns with a precision greater than 6
    (microseconds) also trigger a `NOTICE`, since Postgres caps precisions at
    `timestamp(6)` or `time(6)`.

*   Column types without a PostgreSQL counterpart, including the legacy
    `Object('json')` type that predates ClickHouse `JSON`, trigger an error.
```

doc/pg_clickhouse.md
```diff
@@ -912,32 +924,55 @@ cluster to be restart when the library is updated.
 ## Data Types
 
 pg_clickhouse maps the following ClickHouse data types to PostgreSQL data
```
```suggestion
```

doc/pg_clickhouse.md
```diff
@@ -912,32 +924,55 @@ cluster to be restart when the library is updated.
...
-| JSON       | jsonb, json      |                               |
-| String     | text, bytea      |                               |
-| UInt16     | integer          |                               |
-| UInt32     | bigint           |                               |
-| UInt64     | bigint           | Errors on values > BIGINT max |
-| UInt8      | smallint         |                               |
-| UUID       | uuid             |                               |
+types. [IMPORT FOREIGN SCHEMA](#import-foreign-schema) writes the PostgreSQL
+column shown, carrying its type modifier and array dimensions; [CREATE FOREIGN
+TABLE](#create-foreign-table) may declare other types the value converts into:
```
```suggestion
This table presents the preferred mapping of ClickHouse to PostgreSQL data
types. [IMPORT FOREIGN SCHEMA](#import-foreign-schema) uses these mappings,
and derives the appropriate type modifiers and array dimensions. Use [CREATE
FOREIGN TABLE](#create-foreign-table) to declare alternate PostgreSQL types.
```

doc/pg_clickhouse.md
```diff
@@ -1064,6 +1099,43 @@ But reading them as [BYTEA] will not:
 > As a rule, only use [TEXT] columns for encoded strings and use [BYTEA]
 > columns only for binary data, and never switch between them.
 
+### MAP
+
+PostgreSQL holds no map, so a ClickHouse [Map] column reads into `text[][]`,
+one row of key and value per pair, which is how [IMPORT FOREIGN SCHEMA]
+declares it. An array of that shape inserts, each item parsing into the key or
+value type:
```
```suggestion
PostgreSQL provides no type corresponding to the ClickHouse [Map] type.
pg_clickhouse therefore maps [Map] columns to `text[][]`, with each key pair
as is own array. [IMPORT FOREIGN SCHEMA] uses this mapping.
```

doc/pg_clickhouse.md
```diff
@@ -1064,6 +1099,43 @@ But reading them as [BYTEA] will not:
...
+
+A pair holding other than a key and a value raises, as does an item the key or
+value type cannot take. A `Tuple` column works the same way, taking one array
+of its fields.
+
+> [!NOTE]
+> Inserting a `Map` needs the `binary` driver, which takes the column types
+> from the server. The `http` driver builds them from the PostgreSQL column,
+> which names no `Map`, so ClickHouse rejects the rows.
+
```
theory: Okay man, can we do more copy editing on the slopped text? It's so bad.

```suggestion
Any nested array value containing more than two values triggers an error, as
does an any array incompatible with `TEXTP[][]`.

Similarly, a `Tuple` column maps to a `TEXT[]` array with two values, a key
and a value.

> [!NOTE]
> Inserting a `Map` requires the `binary` driver, which derives the column
> types from the server. The `http` driver builds them from the PostgreSQL
> column, which of course isn't a `Map`.

```

CHANGELOG.md
```diff
@@ -11,6 +11,21 @@ All notable changes to this project will be documented in this file. It uses the
...
+    layers, so `Array(Decimal(12,6))` imports as `numeric(12,6)[]`. It retains up
+    to six digits of `DateTime64(P)` and `Time64(P)` precision, imports `Time`,
+    `Time64`, and geometric types, and maps `FixedString(N)` to unconstrained
+    `text` because ClickHouse counts bytes while PostgreSQL character limits
+    count characters ([#349])
+*   A `Tuple` or `Map` column read into a PostgreSQL array now fills the array
+    with each record's fields rather than a record literal, so `{'k': 'v'}`
+    reads as `{{k,v}}` rather than `{"(k,v)"}`. A composite or `text` column
+    still reads a record. `IMPORT FOREIGN SCHEMA` declares such columns `text[]`
+    and `text[][]`; `binary` driver takes those same shapes on `INSERT`, parsing
```
```suggestion
    and `text[][]`; the `binary` driver accepts those same types on `INSERT`, parsing
```

theory: So, does this work if the `Map` key or value is not a string?

src/pglink.c
```diff
@@ -917,193 +917,213 @@ binary_finalize_insert(void* istate) {
...
+            declaration[len] == '(') {
+            return aggregate_wrappers[i];
+        }
+    }
+    return NULL;
+}
+
+/*
+ * Take one parameter off a ClickHouse parameter list, advancing `params` past it
+ * Return NULL once the list is empty
```
theory: Let's end sentences with periods, eh? Applies for all comments.

src/pglink.c
```diff
@@ -917,193 +917,213 @@ binary_finalize_insert(void* istate) {
...
 
-                *options = lappend(*options, makeInteger(val));
-                *options = lappend(*options, makeString(func));
-            }
+    type         = pgch_pg_type_for(parsed, what);
+    *is_nullable = type.nullable;
 
-            return parse_type(table_name, colname, pos2 + 2, is_nullable, options);
-        }
+    /* Tuple and Map reach the pseudotypes no table column can hold */
```
theory: Not strictly true, since you can use a composite type for Map, IIRC.

src/pglink.c
```diff
@@ -917,193 +917,213 @@ binary_finalize_insert(void* istate) {
...
-    while (str_types_map[i][0] != NULL) {
-        if (strncmp(str_types_map[i][0], typepart, strlen(str_types_map[i][0])) == 0) {
-            return pstrdup(str_types_map[i][1]);
-        }
-        i++;
+    if (type.truncated) {
+        elog(
+            NOTICE,
+            "pg_clickhouse: ClickHouse <%s> exceeds microseconds, %s truncates it for "
+            "column \"%s\"",
```
theory:
```suggestion
            "pg_clickhouse: ClickHouse <column \"%s\"> precision exceeds microseconds (6), %s truncates it for "
```

theory (approved):
Great, mostly editorializing.

CHANGELOG.md
serprex: yes

doc/pg_clickhouse.md
serprex: I agree, haven't put enough effort into the docs side, will find time for that

PR #351, opened 2026-08-19
Remove unnecessary `memset()`s

theory:


serprex (approved):

src/fdw.c
serprex:
```suggestion
    nulls  = palloc0(tupdesc->natts * sizeof(bool));
```
https://github.com/postgres/postgres/blob/cb217c4f174f487009bcba3812bbf23b8d8eba4c/src/backend/utils/sort/tuplestore.c#L783
https://github.com/postgres/postgres/blob/cb217c4f174f487009bcba3812bbf23b8d8eba4c/src/backend/access/common/heaptuple.c#L1415
pg code scans all null entries, but just clearing at start we're good

theory: Done in 674bdf0, thanks.

PR #353, opened 2026-08-25
Add string-backed JSON path pushdown

k-bx:
Compatibility views often expose a ClickHouse String as PostgreSQL json/jsonb through json[b]_in(text::cstring). This teaches json/jsonb extract_path_text pushdown to recognize that exact wrapper and deparse constant non-numeric paths with ClickHouse JSON functions.

The mapping preserves SQL NULL, missing keys, JSON null, empty strings, scalars, objects, nested paths, and invalid-JSON errors. Numeric and empty path components, non-text extraction, and non-constant paths remain local where semantics cannot be mapped safely.

Regression coverage includes Cyrillic ILIKE pushdown and json/jsonb result semantics.

Tested with PostgreSQL 18 and ClickHouse 26.3.12.3:
- full installcheck suite: 39/39 passed
- PostgreSQL 14 compilation
- all pre-commit hooks


src/deparse.c
```diff
@@ -1047,6 +1126,16 @@ foreign_expr_walker(Node* node, foreign_glob_cxt* glob_cxt, ExprTruthCtx ctx) {
                 return false;
             }
 
+            document = serialized_json_document((Expr*)linitial(fe->args));
+            if (document != NULL) {
+                if (cdef->cf_type != CF_JSON_EXTRACT_PATH_TEXT ||
```
theory, line 1093: Why not allow `CF_JSON_EXTRACT_PATH`, too?

src/deparse.c
```diff
@@ -905,6 +909,79 @@ classify_notin_subplan(SubPlan* subplan, PlannerInfo* root, Relids relids) {
...
+serialized_json_document(Expr* expr) {
+    FuncExpr* input;
+    Expr* argument;
+
+    if (!IsA(expr, FuncExpr)) {
+        return NULL;
+    }
+    input = (FuncExpr*)expr;
+    if ((input->funcid != F_JSON_IN && input->funcid != F_JSONB_IN) ||
+        (input->funcresulttype != JSONOID && input->funcresulttype != JSONBOID) ||
```
theory, line 895: Don't `json_in()` and `jsonb_in()` return JSON and JSONB by definition? This line seems redundant.

src/deparse.c
```diff
@@ -905,6 +909,79 @@ classify_notin_subplan(SubPlan* subplan, PlannerInfo* root, Relids relids) {
...
+        list_length(input->args) != 1) {
+        return NULL;
+    }
+
+    argument = (Expr*)linitial(input->args);
+    if (!IsA(argument, CoerceViaIO) ||
+        ((CoerceViaIO*)argument)->resulttype != CSTRINGOID) {
+        return NULL;
```
theory, line 903: `json_in()` and `jsonb_in()` take a `cstring` argument by definition, no? Again seems redundant.

src/deparse.c
```diff
@@ -905,6 +909,79 @@ classify_notin_subplan(SubPlan* subplan, PlannerInfo* root, Relids relids) {
...
+    }
+    argument = ((CoerceViaIO*)argument)->arg;
+    return exprType((Node*)argument) == TEXTOID ? argument : NULL;
```
theory, line 906: We only ever allow TEXT to be converted to a cstring? What about bytea, character, or varchar?

src/deparse.c
```diff
@@ -905,6 +909,79 @@ classify_notin_subplan(SubPlan* subplan, PlannerInfo* root, Relids relids) {
...
+    for (; *cursor != '\0'; cursor++) {
+        if (*cursor < '0' || *cursor > '9') {
+            return false;
+        }
+    }
+    return true;
+}
+
+static bool
+serialized_json_path_is_shippable(Const* path) {
```
theory, line 928: Could use a comment to explain that a numeric value is not usable for ClickHouse JSON path expressions, so this function prevents them from being pushed down.

src/deparse.c
```diff
@@ -4008,9 +4097,88 @@ deparseSubscriptingRef(SubscriptingRef* node, deparse_expr_cxt* context) {
     }
 }
 
+static void
+append_serialized_json_function(
```
theory, line 4062: Please add doc comments for each new function. TIA!

theory (changes requested):
This looks very useful, and simpler than the earlier PRs, thank you!

I rebased and made some tweaks, but have a few more questions about the code and some changes to make.

I also tried using `body::jsonb` instead of `jsonb_in(body::cstring)`; shouldn't that work, as well? Also `CAST(body AS jsonb)`.

PR #355, opened 2026-09-01
Map UInt64 & Int128/UInt128/Int256/UInt256 to numeric

serprex:
ClickHouse/pg_chdb/pull/78

### DIFF
--- a/src/deparse.c
+++ b/src/deparse.c
@@ -1875,6 +1875,12 @@ ch_format_type_extended(Oid type_oid, int32 typemod, uint16 flags) {
         buf = pstrdup("Int64");
         break;
 
+#if PG_VERSION_NUM >= 190000
+    case OID8OID:
+        buf = pstrdup("UInt64");
+        break;
+#endif
+
     case NUMERICOID:
         if (with_typemod) {
             /*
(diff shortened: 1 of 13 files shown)


src/deparse.c
```diff
@@ -1919,6 +1919,12 @@ ch_format_type_extended(Oid type_oid, int32 typemod, uint16 flags) {
         buf = pstrdup("Int64");
         break;
 
+#if PG_VERSION_NUM >= 190000
+    case OID8OID:
+        buf = pstrdup("UInt64");
+        break;
+#endif
```
theory, line 1882: Worth a changelog mention?

theory (approved):

src/deparse.c
serprex: I don't think so, PG19 not even out yet

theory: Does this not mean that we support mapping an OID8 column to UInt64 on Postgres 19? I see no reason not to say so.

serprex: this only supports inserting oid8/xid8 to UInt64, but selecting doesn't work. There's no coercion path for oid8/xid8 from numeric we convert to. We could hack it by having our last effort be to take text output of type & read it to destination type, which for numeric & oid8/xid8 would work

theory: Ah, yes, I see.

PR #356, opened 2026-09-02
Drop support for Postgres 13

theory:
Remove a ton of `#if` directives, including some quite old ones for Postgres 12! Remove Postgres 13-specific expected test output files and update `result_map.txt`.


serprex (approved):

serprex:
#357 missed a spot

PR #357, opened 2026-09-03
Remove PG13 compat

serprex:


src/fdw.c
```diff
@@ -27,9 +27,6 @@
 #include "utils/palloc.h"
 #include "utils/rel.h"
 #include "utils/tuplestore.h"
-#if PG_VERSION_NUM >= 140000
-#include "optimizer/appendinfo.h"
```
theory, line 31: Erm, don't we need to keep this?

theory (approved):

src/fdw.c
serprex: unused

PR #360, opened 2026-09-06
Add pgcrypto digest and SHA function pushdown

### LINKED ISSUE #21: Map Postgres Hashing Functions to ClickHouse Functions
Both core functions and [pgcrypto](https://www.postgresql.org/docs/current/pgcrypto.html)'s `digest` function:

| PostgresSQL | ClickHouse |
|-------------|------------|
| `md5(foo)`  | `MD5(foo)` |
| `sha224(foo)` | `SHA224(foo)` |
| `sha256(foo)` | `SHA256(foo)` |
| `sha384(foo)` | `SHA384(foo)` |
| `sha512(foo)` | `SHA512(foo)` |
| `digest(foo, 'md5')` | `MD5(foo)` |
| `digest(foo, 'sha224')` | `SHA224(foo)` |
| `digest(foo, 'sha256')` | `SHA256(foo)` |
| `digest(foo, 'sha384')` | `SHA384(foo)` |
| `digest(foo, 'sha512')` | `SHA512(foo)` |

And potentially others that [pgcrypto](https://www.postgresql.org/docs/current/pgcrypto.html) automatically picks up from OpenSSL.

sivagirish81:
Added pushdown support for PostgreSQL’s `sha224()`, `sha256()`, `sha384()`, and `sha512()` functions, along with both `pgcrypto.digest(text, text)` and `pgcrypto.digest(bytea, text)`.

`digest()` is pushed down when its algorithm is a recognized, non-`NULL` constant, mapping `md5`, `sha1`, `sha224`, `sha256`, `sha384`, and `sha512` to the corresponding ClickHouse functions. Dynamic, `NULL`, and unsupported algorithms remain local to PostgreSQL.

Added regression coverage for pushdown behavior and result validation, including `bytea` values containing NUL and non-UTF-8 bytes, and documented the new mappings.

Closes #21


test/sql/hashing.sql
```diff
@@ -0,0 +1,132 @@
...
+BEGIN
+    FOR i IN 1..array_length(pg_names, 1) LOOP
+        EXECUTE format(
+            'EXPLAIN (VERBOSE, FORMAT JSON) '
+            'SELECT %I(binary_data) AS h FROM hash_inputs GROUP BY h',
+            pg_names[i]
+        ) INTO output;
+        remote_sql := jsonb_path_query_first(output, '$[0].Plan')->>'Remote SQL';
+        RAISE NOTICE '% PUSHED DOWN: %', pg_names[i],
+            strpos(remote_sql, ch_names[i] || '(binary_data)') > 0;
```
theory: Wouldn't this also return true if it wasn't pushed down?

test/sql/hashing.sql
```diff
@@ -0,0 +1,132 @@
...
+DECLARE
+    pg_names text[] := ARRAY['sha224', 'sha256', 'sha384', 'sha512'];
+    ch_names text[] := ARRAY['SHA224', 'SHA256', 'SHA384', 'SHA512'];
+    digest_names text[] := ARRAY['md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512'];
+    digest_ch_names text[] :=
+        ARRAY['MD5', 'SHA1', 'SHA224', 'SHA256', 'SHA384', 'SHA512'];
+    output jsonb;
+    remote_sql text;
+BEGIN
+    FOR i IN 1..array_length(pg_names, 1) LOOP
```
theory: Prefer [array looping](https://www.postgresql.org/docs/current/plpgsql-control-structures.html#PLPGSQL-FOREACH-ARRAY).

test/sql/hashing.sql
```diff
@@ -0,0 +1,132 @@
...
+    text_data text,
+    binary_data bytea,
+    algorithm text
+) SERVER hashing_loopback OPTIONS (table_name 'inputs');
+
+-- Check core SHA and pgcrypto digest pushdown without depending on EXPLAIN's
+-- text formatting, which differs across supported PostgreSQL versions.
+\unset ECHO
+DO $$
+DECLARE
```
theory: Is there so much variation in `EXPLAIN` output? Would be nice to see and compare the actual plans.

theory (changes requested):
Looks fairly straightforward. If you'd be so kind as to respond to this feedback, I'll take the next iteration for a more thorough spin. Thank you.

theory:
Also, please add a `CHANGELOG.md` item and link to this PR. TY!

test/sql/hashing.sql
sivagirish81: There can be some `EXPLAIN` output differences across PostgreSQL major versions, which is why I initially avoided comparing the full output. These plans are simple enough that the important details should remain stable, though, and showing them makes the pushdown behavior much clearer. I’ve replaced the PL/pgSQL assertions with explicit `EXPLAIN (VERBOSE, COSTS OFF)` statements and regenerated the expected output with the actual plans.

test/sql/hashing.sql
sivagirish81: Fixed. Switched to explicit `EXPLAIN` statements, so the array loop is no longer needed.

test/sql/hashing.sql
sivagirish81: Addressed. The plan output now directly shows whether each function appears in `Remote SQL` or remains in the local PostgreSQL filter.

theory (approved):
Thank you!

PR #361, opened 2026-09-14
Add the `encoding_check` server option

theory:
With 924964d, pg-clickhouse-c raises an error when fetching text or JSON data that doesn't adhere to the database encoding. This changes the previous behavior, which simply let invalidly-encoded text pass through. As a result, existing installations may start to experience errors after upgrading.

Upgrade pg-clickhouse-c again to get its new `encoding_check` attribute of `pgch_reader`. Then add the `encoding_check` server option. This lets one configure server-specific handling of encoding errors: truncate, replace, or remove. Teach `chfdw_extract_options()` how to parse and validate its values.

Passing this setting proves a bit tricky, however; it must be parsed external to connection parameters, since it's not a client option. However we save it in the `ch_connection` struct to allow access anywhere a query will be executed.

Add the `encoding_check` field to the `ch_query` struct and set it from the `ch_connection` after the construction of the connection and before each call to `chfdw_cursor_open()`, which now simply assigns it to the `pgch_reader` before passing it off to pg-clickhouse-c to use.

Add a test, `test/sql/encoding.sql`, to validate the behavior of the option. It runs only for the UTF-8 or SQL_ASCII encodings, and simply exits for any other encoding.

Alter `Makefile` to allow `REGRESS_OPTS` to be set on the command-line, then add a step to the Postgres workflow to test with `SQL_ASCII`.

While at it:

*   Ignore some addition files for editors and macOS
*   Upgrade codespell and teach it to ignore `*.out` files
*   Reformat some struct declarations to name each field on its own line


src/option.c
```diff
@@ -274,6 +291,23 @@ parse_min_tls_version(const char* val, tls_version* out) {
     return true;
 }
 
+/* Map an encoding_check option value to pgch_encoding_check. */
+static bool
+parse_encoding_check(const char* val, pgch_encoding_check* out) {
+    if (pg_strcasecmp(val, "replace") == 0) {
+        *out = CHC_ENC_REPLACE;
+    } else if (strcmp(val, "remove") == 0) {
```
serprex: why inconsistent strcmp?

theory: Copied from parse_min_tls_version

serprex: `parse_min_tls_version` always uses `pg_strcasecmp`

theory: Bah, only read the first one, sorry.

theory: Fixed and tested in 6c91df6.

serprex (approved):

PR #367, opened 2026-09-22
CH 26.9 map cardinality to arrayFlattenedLength

serprex:
https://clickhouse.com/docs/reference/functions/regular-functions/array-functions#arrayFlattenedLength


src/deparse.c
```diff
@@ -1430,31 +1430,28 @@ foreign_expr_walker(Node* node, foreign_glob_cxt* glob_cxt, ExprTruthCtx ctx) {
...
- * rel is accessed on behalf of another user, e.g. through a view --- wins
- * over the invoking user. Returns NULL instead of erroring when no mapping
- * exists, so the version gate degrades to "no pushdown" rather than failing a
- * query (or a bare EXPLAIN) at plan time.
+ * True when the server that scans foreignrel is at least major.minor. Probes
+ * with the user mapping the executor will use (see clickhouseBeginForeignScan):
+ * the RTE's checkAsUser --- set when the rel is accessed on behalf of another
+ * user, e.g. through a view --- wins over the invoking user. Returns false
+ * instead of erroring when no mapping exists, so the gate degrades to "no
+ * pushdown" rather than failing a query (or a bare EXPLAIN) at plan time.
```
```suggestion
 * True when the server that scans foreignrel is at least major.minor. Probes
 * with the user mapping that clickhouseBeginForeignScan() sets up for the
 * executor. Returns false instead of erroring when no mapping exists, so the
 * gate degrades to "no pushdown" rather than failing a query (or a bare
 * EXPLAIN) at plan time.
```

theory (approved):

PR #371, opened 2026-09-22
Test Nested mapping; document composite mapping

theory:
Add `test/sql/nested.sql` to comprehensively test the behaviors of `Nested` column mapping with `flatten_nested=1` and `flatten_nested=0`, as well as the behaviors with `text[][]` and composite type mapping in the latter case.

Expand the `Map` documentation into documentation for all of the ClickHouse composite types: `Array`, `Map`, `Tuple`, and `Nested`, with copious examples. Move and copy-edit the `Manual Type Mappings` section to follow these docs, and test the example in `test/sql/nested.sql`, as well.


theory:
Some tests currently fail:

* Nested array access pushdown:

    ```sql
    SELECT visit_id FROM nest_bin.visits WHERE goals[1][1] = '1';
    ```

*   `INSERT`ing a Nested value mapped to a composite type:

    ```sql
    VALUES (3, 3, ARRAY[row(5, 'jj'), row(6, 'zz')]::goal_type[]);
    INSERT INTO nest_bin.visits
    ```

    Probably should be fixable for the binary driver at least.

PR #374, opened 2026-09-23
update pg-clickhouse-c, improves integer/interval coercion

serprex:


test/expected/interval_int.out
theory, line 1: Maybe fix #364?

theory (changes requested):
Also changelog entry and doc changes.

test/expected/interval_int.out
serprex: #364 touches on a few different things like driver specific changes being needed, whereas this is trying to just be a dependency update

theory: Okay. I'll rebase it after this, then.

theory (approved):

